PR #10, opened 2019-03-03
fix: Fix compatibility with Tornado 6.0

jbms:
When set_nodelay is called from the WebsocketHandler.open method in Tornado 6.0,
it results in an assertion failure.  This change introduces a workaround.


jbms:
The related Tornado bug is https://github.com/tornadoweb/tornado/issues/2611

Issue #14, opened 2019-06-01
HTTPS


witten:
This is a feature request to support HTTPS, not in beancount-import directly, but in any reverse proxy that's wrapping it. Let me back up a bit: My use case is that I'd like to securely host beancount-import on an HTTPS web server so that both I and my partner can collaborate on processing our shared pending transactions. I realize that the typical use case is to run beancount-import locally and then connect to localhost via HTTP, but I'd prefer to run it on an actual web server if possible to aide in collaboration.

I'm wrapping beancount-import with nginx to provide the HTTPS termination, and then proxy passing requests on to the local beancount-import web server. The problem appears to be that `frontend/server_connection.ts` includes `ws://`, hard-coded:

```javascript
"ws://" + window.location.host + "/" + secretKey + "/websocket"
```

This results in a `SecurityError: The operation is insecure.` console error in Firefox when using an HTTPS URL. This is because Firefox prevents HTTPS web pages from accessing non-secured web sockets. So in this case, what I need is for the front-end to access the web socket via a secured `wss://` URL. I imagine one way to support this properly would be to fork on `ws://` or `wss://` based on the current page's protocol (HTTP or HTTPS, respectively).

Before I try to do this, I wanted to get a read on whether this is something you'd entertain. Is this use case too far outside the bounds of how you see beancount-import being used? Thanks!

jbms:
That certainly makes sense --- I've thought about doing something similar for the same reason, but haven't attempted it.   So changes to make this work better would certainly be welcome.  If all that needs to be done is to swap wss:// for ws://, then even better.  I think more generally some other changes may be helpful to make the beancount-import webserver more convenient as a persistent, multi-user service.  For example, currently while it will reload the journal automatically there is no way to reload the import data, and failure to parse the import data often leads to terminating the program.

witten:
Took a stab at this with the linked PR.

Agreed that some other changes may be necessary to better support multi-user. I can look into them separately as I run into each one.

PR #15, opened 2019-06-01
HTTPS / TLS websocket support (#14)

witten:
See https://github.com/jbms/beancount-import/issues/14 for the motivation / background.

### Changes

This changes two things with the URL used to connect from the front-end to the websocket:

1. Uses `ws://` when on a HTTP page, and `wss://` on an HTTPS page.
2. Supports connecting to a `/prefix/path/` in case the whole application is running there instead of at the root of a domain. This is necessary if you want to, say, run `beancount-import` on an existing non-localhost domain along with other applications.

### Testing done

1. Ran `beancount-import` locally on localhost, connecting directly to the Tornado server (well, passing through Docker).  The idea was to confirm that these changes did not regress the localhost non-HTTPS case. Observed that `beancount-import` still appeared to load and work.
2. Ran `beancount-import` on a server behind nginx. This was to confirm that the new HTTPS-supporting functionality works. Observed that `beancount-import` appeared to load and work via an HTTPS URL.

### Other notes

In case anyone's interested, here's the nginx location config I used to front for `beancount-import`:

```
location /some/url/prefix/ {
    proxy_pass_header Server;
    proxy_set_header Host $http_host;
    proxy_redirect off;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Scheme $scheme;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "Upgrade";
    proxy_pass http://localhost:8101/;
}
```


jbms:
One thing that could be useful is to add the nginx configuration you have in the notes to the README or other documentation file so that it is easier to find.

witten:
Good idea! Just added this. Not sure if it's the right spot in the docs, so I'd be happy to move it / revise it as necessary.

jbms:
Thanks!

Issue #17, opened 2019-06-04
High-ish idle CPU usage


witten:
When I've got Beancount-import running, I notice that it appears to idle at around 3% (on an arbitrary machine) as viewed in `top`. This is when the front-end is not open in a browser. Just based on a cursory code inspection, my guess is that this may be due to the check for journal modifications every 100 milliseconds in `webserver.py`:

```
       self.check_modification_timer = tornado.ioloop.PeriodicCallback(
            self._check_modification, 100)
```

Is there any reason this check is so frequent? Just to get fast updates to the user?

jbms:
There could be an option to check less frequently, but in part checking frequently is useful to avoid concurrent modification issues --- if you edit the file manually from a text editor, it is good for beancount-import to know about the changes before attempting to re-write a modified version of the file.  There is a separate check just before it actually overwrites any existing journal file, but I believe that just leads to the program terminating if that check fails.  The best solution would be to use inotify on Linux and similar APIs on other operating systems to monitor changes without polling.

witten:
Got it, thanks for the explanation. I'll have a look at the inotify approach.

PR #19, opened 2019-06-06
Use inotify/equivalent for detecting journal modifications (#17).

witten:
### Background

See https://github.com/jbms/beancount-import/issues/17. The short version is that this change reduces Beancount-import's CPU usage when idle.

### Changes

* Take a new dependency on [Watchdog](https://github.com/gorakhargosh/watchdog), a Python library that wraps [inotify](https://en.wikipedia.org/wiki/Inotify) filesystem change notifications on Linux, and comparable APIs on other OSes.
* Create a Watchdog observer and handler that simply watches for any filesystem events in the directory that Beancount-import's journal input file resides in. When any such events are received, poke the existing Beancount-import journal check modification function to do its thing.
* Rip out the periodic callback now that we're using Watchdog/inotify to find out about filesystem changes.
* Disable Tornado's debug mode unless the existing `--debug` parameter is passed to Beancount-import. More about that below!

### Performance impact

* As a baseline, I tested without these changes. After initialization, an idle Beancount-import took 3-4% CPU on a particular machine.
* After making all of the changes above, CPU dropped down to near-0%.
* In trying to isolate which changes had the most impact, I tried enabling only the journal change detection improvements, while leaving Tornado debug mode on. CPU consumption was around 3%!
* Then I tried reverting the journal change detection improvements, and just turning off Tornado debug mode. CPU consumption was around 0.7%!

Take-away: Most of the CPU consumption was actually Tornado's debug mode. Why? Well, enabling debug mode also enables autoreload, which polls all source files for changes repeatedly so as to better support the development use case of autoreloading the web server. (I confirmed it was indeed autoreload by looking at Python profiling output.)

Even though the journal change detection improvements don't have a huge impact compared to disabling the debug mode, I personally still think both are worth it. Happy to hear other opinions though!

### Testing

* Touched a journal file and ensured that it triggered a refresh. Observed this in debug output.
* Modified a journal file with vim and ensured that it triggered a refresh. Observed this both in debug output and in the front-end. I wanted to make sure the Watchdog approach worked with vim, because vim does some tricky shit with file replacement instead of editing the file in place.
* Touched an unrelated file in the same directory and ensured that it triggered a refresh. Observed this in debug output.

### Caveats

* Since the Watchdog-based changed detection only operates on the directory containing the journal input file, changes to files in unrelated directories (e.g. via an include) won't trigger a refresh. Let me know if you think that's a problem.
* The change detection triggers when *anything* changes in the monitored directory. Random file deletes/adds, directory changes, etc. That doesn't seem to be a problem, because all it does is poke the existing Beancount-import code to stat journal files. So if it's a little trigger happy, that's probably fine.

### DIFF
--- a/beancount_import/webserver.py
+++ b/beancount_import/webserver.py
@@ -26,6 +26,9 @@ from beancount.core.data import Transaction, Posting
 from beancount.core.number import MISSING, Decimal
 import beancount.parser.printer
 
+import watchdog.events
+import watchdog.observers
+
 from . import reconcile
 
 from . import training
@@ -393,6 +396,15 @@ class GetFileHandler(tornado.web.RequestHandler):
             self.finish('File not found')
 
 
+class JournalModificationHandler(watchdog.events.FileSystemEventHandler):
+    def __init__(self, application):
+        super(JournalModificationHandler, self).__init__()
+        self.application = application
+
+    def on_any_event(self, event):
+        self.application.ioloop.add_callback(self.application.check_modification)
+
+
 class Application(tornado.web.Application):
     def __init__(self, args, ioloop, **kwargs):
 
@@ -431,8 +443,10 @@ class Application(tornado.web.Application):
             options=vars(args))
         self.reset()
 
-        self.check_modification_timer = tornado.ioloop.PeriodicCallback(
-            self._check_modification, 100)
+        self.check_modification_timer = watchdog.observers.Observer()
+        self.check_modification_timer.schedule(
+            JournalModificationHandler(self),
+            os.path.dirname(os.path.realpath(args.journal_input)))
         self.check_modification_timer.start()
 
     def next_generation(self):
@@ -453,7 +467,7 @@ class Application(tornado.web.Application):
                 except:
                     traceback.print_exc()
 
-    def _check_modification(self):
+    def check_modification(self):
         if self.reconciler.loaded_future.done():
             loaded_reconciler = self.reconciler.loaded_future.result()
             modified_filenames = loaded_reconciler.editor.check_any_journal_modification(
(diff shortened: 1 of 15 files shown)


jbms:
This is great, thanks!

Potentially it would be nice to make it handle the case of journal files not all being in the same directory by adding all directories containing journal files (and then after the journal reloads, updating the list of directories again).

If you feel like writing a unit test for the reload detection, that would also be particularly awesome, though not necessary considering that a lot of existing functionality also lacks test coverage.

witten:
Good call. I was able to change up the file monitoring to reflect all journal file paths, even via includes, and even after updates as you suggest.

Manual testing:
 * Added a debugging print of the current journal paths right before registering monitoring for each path.
 * Started Beancount-import with all journal files in the same directory: `/data`. Observed via debugging print that it was monitoring just `/data`.
 * Added an empty `/data/tmp/test.beancount` file. Observed no changes.
 * Added an include in `/data/journal.beancount` to pull in `/data/tmp/test.beancount`. Observed via debugging print that now `/data` and `/data/tmp` were being monitored.
 * Added a bogus `balance` entry to `/data/tmp/test.beancount`. Observed a refresh and saw a corresponding balance error in the front-end.
 * Removed the bogus `balance` entry. Observed a refresh and saw the error go away.
 * Removed the include. Observed via debugging print that now just `/data` was being monitored again.
 * Removed `/data/tmp/test.beancount`. Observed no change.

I'll look into adding some automated testing to this PR when I get a chance!

witten:
Well, it's not pretty, but there's now an automated test similar to the above manual steps for the journal reload detection. Ready for re-review!

jbms:
Thanks!

PR #31, opened 2019-10-04
feature: support filtering by year for ultipro_google and stockplanconnect sources

jktomer:
I organize most of my source documents by year, e.g. .../Pay
Statements/2018/2018-01-02.statement.pdf or .../Amazon
Invoices/2019/123-45678-9012.html.

For most sources, I can either use a recursive glob to pick up all the files
in the tree, or create one Source per year, and everything works fine. But for
stockplanconnect and ultipro_google, which expect to be able to translate from
a transaction in the journal to a source path based only on the document ID,
this doesn't work; if I use a recursive glob, every journal entry gets
declared invalid because the statements don't have the expected path, and if I
use one source per year-specific subdirectory, every entry ends up being
declared invalid by the sources that only expect to find documents from other
years.

This change allows the use of a separate source per year for
directory-per-year layouts, by adding a "year" parameter to these two sources,
and ignoring any journal entry from a date outside the configured year if
there is one.

### DIFF
--- a/beancount_import/source/stockplanconnect.py
+++ b/beancount_import/source/stockplanconnect.py
@@ -61,7 +61,7 @@ from beancount.core.amount import sub as amount_sub
 from beancount.core.position import CostSpec
 from beancount.core.number import D, ZERO
 from beancount.core.number import MISSING
-from beancount_import.posting_date import POSTING_DATE_KEY
+from beancount_import.posting_date import POSTING_DATE_KEY, get_posting_date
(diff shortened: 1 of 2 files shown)


jbms:
This seems okay, though having one source per year does mean having to go through the entire transaction list for each year in certain steps.  You could instead change the source to support per-year directories.  Either way it is okay for now, though.

jktomer:
I actually do like supporting per-year directories within the source a bit better; let me take a crack at that.

PR #33, opened 2019-10-05
fix(healthequity): Match fund transactions by price, not cost.

jktomer:
This solves two problems:
* Sell transactions have no cost specified, but were being matched incorrectly
  because get_key_from_raw_entry gave a cost of None, but get_key_from_posting
  gave a cost of (MISSING, MISSING). This could be fixed by using None instead
  of (MISSING, MISSING) in get_key_from_posting, but that seems strictly worse
  than using price instead, as the latter is highly likely to actually
  match (only) a correct journal entry.

* Buy transactions may be unbalanced as imported by default due to
  rounding (on occasion I've seen HealthEquity round differently than
  beancount wants to, i.e. truncating rather than rounding to the nearest
  thousandth-share on buys). If the user fixes them by manually editing the
  journal to change per-share cost into total cost, they will no longer match
  the input properly. Using the price for matching instead fixes this: it's OK
  for the price not to exactly equal the per-share cost, so there's no reason
  for the user to modify it.

Because we now match investment transactions to existing entries by price
rather than cost, we now also specify price (in addition to cost) for buy
transactions, so that we can continue to match buys. There seems to be no
downside to doing this; the fact that we bought a security at a given price on
a given day strongly suggests that this was a correct price for it on that
day. (Alternatively, we could match transactions to the journal by
price-or-cost, which might be helpful for backward compatibility.)


jbms:
Thanks for this fix.  As you can probably guess I don't have any sell transactions which is why this went unnoticed.  I always use the implicit_prices plugin so the costs contribute to prices anyway, but your argument of being able to fix rounding errors by editing the cost but not the price makes sense.

I think it would be useful to match buy transactions by price first, but use cost if there is no price, for backward compatibility.

jktomer:
> I think it would be useful to match buy transactions by price first, but use cost if there is no price, for backward compatibility.

Done, and added a test for that case to test_matching. PTAL.

jbms:
Thanks!

Issue #26, opened 2019-09-20
Amazon invoices have gotten much less useful :(


jktomer:
Amazon invoices have changed recently, and no longer show any summaries on a per-shipment basis. This means pre-/post-tax adjustments and sales tax can no longer be calculated per shipment :cry:

The changes were even made retroactively; see e.g. this example of a previously-downloaded invoice vs. the newly-generated invoice for the same order.

[old-good-invoice.pdf](https://github.com/jbms/beancount-import/files/3634456/old-good-invoice.pdf)
[new-bad-invoice.pdf](https://github.com/jbms/beancount-import/files/3634455/new-bad-invoice.pdf)

I have complained to Amazon but I doubt it's going to change anything, so I plan to fix up the parser to handle the eviscerated statements this weekend; mainly opening this issue to have something to reference from the pull request.

Zburatorul:
@jktomer, have you worked on the Amazon importer since?

jktomer:
> @jktomer, have you worked on the Amazon importer since?

I have a few tweaks in my local repo that I need to clean up and upstream.

Zburatorul:
@jktomer, can I help with anything?

Zburatorul:
@jktomer, would love to make use of your improvements to this source.

PR #37, opened 2019-10-12
feat(amazon): handle invoices with no per-shipment details

jktomer:
Amazon has recently changed their invoice format to remove the per-shipment
totals tables (see issue #26 ). This change loosens parsing to handle such
invoices, without breaking parsing of older invoices.

Pre-tax adjustments and sales tax, if not present in the shipment tables, are
added at the per-invoice level instead. If there is only one shipment in the
transaction, they continue to be grouped with the lone shipment; otherwise,
like unrecognized post-tax adjustments, each is ungrouped.


jbms:
Thanks for fixing this!

Issue #45, opened 2020-01-21
Tornado - async IO throwing NotImplementedError


TiraelSedai:
Used to work fine until I reinstalled Windows. I used to have 3.7.something Python. It could be some other dependency.

Steps to reproduce:

1. Fresh windows
2. Install VS build tools
3. Get Python 3.8.1
4. python -m pip install --upgrade pip
5. pip install beancount-import
6. Run beancount-import

```
Traceback (most recent call last):
  File "C:\Users\Tirae\OneDrive\Ledger\run_win.py", line 33, in <module>
    run_reconcile(sys.argv[1:])
  File "C:\Users\Tirae\OneDrive\Ledger\run_win.py", line 19, in run_reconcile
    beancount_import.webserver.main(
  File "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\site-packages\beancount_import\webserver.py", line 737, in main
    http_server.add_sockets(sockets)
  File "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\site-packages\tornado\tcpserver.py", line 165, in add_sockets
    self._handlers[sock.fileno()] = add_accept_handler(
  File "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\site-packages\tornado\netutil.py", line 279, in add_accept_handler
    io_loop.add_handler(sock, accept_handler, IOLoop.READ)
  File "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\site-packages\tornado\platform\asyncio.py", line 99, in add_handler
    self.asyncio_loop.add_reader(fd, self._handle_events, fd, IOLoop.READ)
  File "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\asyncio\events.py", line 501, in add_reader
    raise NotImplementedError
NotImplementedError
```

TiraelSedai:
Temporal fix:
Go to tornado/platform/asyncio.py, in my case it's "C:\Users\Tirae\AppData\Local\Programs\Python\Python38-32\lib\site-packages\tornado\platform\asyncio.py"

Add
import sys

and after imports:

if sys.platform == 'win32':
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

TiraelSedai:
[...]

PR #46, opened 2020-01-25
Fix Windows + Python 3.8 changes to Tornado async IO loop

TiraelSedai:
https://github.com/jbms/beancount-import/issues/45

Basically implement what's been discussed in issue thread.


jbms:
Thanks!

PR #48, opened 2020-01-27
Fix ofx assert for Fidelity. Fixes #47

### LINKED ISSUE #47: Failed assertion for Fidelity accounts
Importing ofx files from my Fidelity 401k account for BUY transactions is problematic. 

In ofx.py:1040 there's this assertion:
if raw.trantype in STOCK_BUY_SELL_TYPES:
                assert abs(total + fee_total +
                           (units * unitprice)) < TOLERANCE, abs(
                               total + fee_total + (units * unitprice))

The unitprice is not "correct" in the ofx file, and tolerance is not met. units and total is correct and they match the statement.

The cost_spec is computed on line 933, using unitprice, before we determine if a fee was charged. 

What I think should happen is that total and fee_total should be computed before the cost_spec initialization, and instead of using "number_per", we should use "number_total" with -(total+fee_total) as its value, and ignore potentially problematic unitprice values.

scanta2:
Fix ofx assert for Fidelity, handle ofx with negative unit price, fix internal account transfers.

No time to run tests.....

### DIFF
--- a/beancount_import/source/ofx.py
+++ b/beancount_import/source/ofx.py
@@ -948,8 +948,11 @@ class ParsedOfxStatement(object):
                     # Internal transfer, i.e. from after-tax to roth
                     continue
                 else:
+                    number_per_fix = unitprice
+                    if abs(total + fee_total + (units * unitprice)) >= TOLERANCE:
+                    	number_per_fix = normalize_fraction((abs(total)-abs(fee_total))/units)
                     cost_spec = CostSpec(
-                        number_per=(abs(total)-abs(fee_total))/units,
+                        number_per=number_per_fix,
                         number_total=None,
                         currency=self.currency,
                         date=None,
--- a/beancount_import/source/ofx.py
+++ b/beancount_import/source/ofx.py
@@ -842,6 +842,21 @@ class ParsedOfxStatement(object):
                 if stripped_checknum:
                     posting_meta[CHECK_KEY] = D(stripped_checknum)
 
+            fee_total = ZERO
+            for fee_key in ['fees', 'commission']:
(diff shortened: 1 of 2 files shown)


jbms:
Thanks for your fix, and I'm really sorry for not getting back sooner.

A few comments:

The test failures are not too serious.  There are two types of issues I spotted:

https://travis-ci.com/github/jbms/beancount-import/jobs/280640548#L826

Here, the calculation of the unitprice results in an excessive number of digits.  While it doesn't really matter that much, I think it would be better to retain the original unitprice if the OFX file specifies a valid one.  Basically, replace the existing assertion that was failing for you with an if statement.  If the original unitprice is within the tolerance, just use it.  Otherwise, calculate a unitprice but add a warning in the form of e.g. txn.meta['ofx_error'] = 'unitprice specified in file ...'

https://travis-ci.com/github/jbms/beancount-import/jobs/280640548#L1323

This failure is just due to the change in the order in which postings are added.  That could be fixed, or the golden file could just be regenerated by running the tests with

BEANCOUNT_IMPORT_GENERATE_GOLDEN_TESTDATA=1

Aside from those minor issues, it would be extremely helpful if you are able to provide a sanitized ofx file as a test case that exhibits this problem.  For example, with a test case I could easily make those changes I suggested above myself, and verify that it is still solves your issue.  That also ensures that this doesn't accidentally get broken in the future by later changes.

jbms:
Note: This relates to #47.

scanta2:
@jbms  All Linux tests now pass. I have rebaselined the order of the transactions in your test harness. Have a look please. Windows tests fail because they are looking for the Python 3.8 interpreter.

jbms:
Thanks

Issue #54, opened 2020-04-11
Entries with 0 unitprice


stin7:
A Vanguard OFX I'd like to import has entries where the UNITPRICE is 0 (which seems like a bug on their part):

```
<REINVEST>
    <INVTRAN>
        <FITID>XXX</FITID>
        <DTTRADE>20190614160000.000[-5:EST]</DTTRADE>
        <DTSETTLE>20190614160000.000[-5:EST]</DTSETTLE>
        <MEMO>DIVIDEND REINVEST</MEMO>
    </INVTRAN>
    <SECID>
        <UNIQUEID>XXX</UNIQUEID>
        <UNIQUEIDTYPE>CUSIP</UNIQUEIDTYPE>
    </SECID>
    <INCOMETYPE>DIV</INCOMETYPE>
    <TOTAL>-543.95</TOTAL>
    <SUBACCTSEC>CASH</SUBACCTSEC>
    <UNITS>7.599</UNITS>
    <UNITPRICE>0.0</UNITPRICE>
</REINVEST>
```

Which results in errors like this:

```
Traceback (most recent call last):
  File "miniconda3/lib/python3.6/site-packages/beancount_import/webserver.py", line 493, in _handle_reconciler_loaded
    loaded_reconciler = loaded_future.result()
  File "miniconda3/lib/python3.6/concurrent/futures/_base.py", line 398, in result
    return self.__get_result()
  File "miniconda3/lib/python3.6/concurrent/futures/_base.py", line 357, in __get_result
    raise self._exception
  File "miniconda3/lib/python3.6/site-packages/beancount_import/thread_helpers.py", line 13, in wrapper
    f.set_result(fn(*args, **kwargs))
  File "miniconda3/lib/python3.6/site-packages/beancount_import/reconcile.py", line 396, in __init__
    all_source_results = self._prepare_sources()
  File "miniconda3/lib/python3.6/site-packages/beancount_import/reconcile.py", line 515, in _prepare_sources
    source.prepare(self.editor, source_results)
  File "miniconda3/lib/python3.6/site-packages/beancount_import/source/ofx.py", line 1394, in prepare
    state.get_accounts_and_entries()
  File "miniconda3/lib/python3.6/site-packages/beancount_import/source/ofx.py", line 1249, in get_accounts_and_entries
    statement.get_entries(self)
  File "miniconda3/lib/python3.6/site-packages/beancount_import/source/ofx.py", line 1043, in get_entries
[...]

PR #55, opened 2020-04-11
Zero unitprice

stin7:
Workaround for Issue #54


jbms:
Thanks for this fix.  It looks like this addresses a similar issue to #47 (see the discussion in particular).  Since I believe UNITPRICE is required according to the OFX spec (though I may be misreading it), it probably makes sense to calculate it if it is missing or inaccurate, but to include a warning via txn.meta['ofx_error'] = '...' to notify the user.

Issue #23, opened 2019-07-21
Import from bean-extract / beancount.ingest?


wadabum:
So, I´ve spend hours writing a bean-extract compatible parser using `beancount.ingest.importer.ImporterProtocol` 

## **Can one somehow string those things together?**

----

- I do NOT mean parsing a written-our .beancount file,
but "run bean-extract" and use that result as data_source - similar to Fava?

something along the lines of the usual:
```python3
# beancount.config
from importers import my_importer
configured_my_importer = my_importer.Importer(
  currency='USD',
)
CONFIG = [
  configured_my_importer,
]
# ....
# ....
# ....
    data_sources = [
        dict(
            module='beancount_import.source.beancount-ingest',
            class=configured_my_importer,
        )
# or
    data_sources = [
        dict(
            module='beancount_import.source.beancount-extract',
            config='path/to/beancount.config', # parse&read CONFIG from there
            dirs='documents/',
        )
```
or however that´s gonna map out... ;-)

(think that could proxy-close #18 as well...)

jbms:
There isn't any existing way to do this, but some sort of convergence with beancount.ingest is something I'm very interested in exploring.  However, there are some challenges to overcome:

1. beancount.ingest is based on generating partial transactions (that don't necessarily balance) with missing postings.  beancount-import requires full transactions, with Expenses:FIXME postings to indicate unknown accounts (a single Expenses:FIXME posting is allowed to correspond to multiple unknown postings).  This is simple to handle, though --- just generate Expenses:FIXME postings to balance the transaction.

2. With just step 1, you could have a partial integration with beancount-import, but to get the automatic tracking of which transactions have already been imported, and automatic classification of unknown accounts, the beancount-import data source needs to provide more information than is available in the beancount.ingest model.

dumbPy:
How about exporting this required info in meta?
I am currently exporting transaction ref/ cheque No. and jumbled description into meta.
ML models that can use this to predict postings and use your awesome UI for manual corrections would be great. currently exploring [smart_importer](https://github.com/beancount/smart_importer) but it doesn't have a way to fix interactively like this project.
I too want this to work.

jbms:
If you specified which metadata fields indicate a unique transaction id, and which metadata fields to use for automatic classification,  then I think it would work.

dumbPy:
Do you have any plans for defining API for the UI part?

[...]

PR #60, opened 2020-07-27
Towards making smart-importer and beancount-import more cooperative.

Zburatorul:
Introduced two new methods similar to those present in training.py and new logic.

beancount-import will now look for `new_account` and `unconfirmed_account` meta fields in each posting, and understand their presence to mean that the account names of these postings are tentative, potentially new. It will then work in a passthrough mode, offering them for import/modifications.

I don't expect this to actually be merged, but rather to serve as a first draft implementation of the proposal from #23.


jbms:
Thanks for working on this!  It looks like you are on the right track.  When you have something working, can you add a test?

dumbPy:
smart-importer wraps the `ImporterProtocol.extract` and modifies the returned `imported_entries`.  
`existing_entries` and `imported_entries` are `List[Transaction]`. Importers generally write the source posting, and target posting is predicted by smart-importer. deduplication is optional and I see that beancount-import does it well, specially from manually entered transactions and I think this would be a better way to go about it.

We will just need to write a generic `ImporterSource(Source)` subclass that takes user defined importers from the config (possibly wrapped in smart-importer), and if there's a single posting, add `Expense:FIXME`. Might need more things, but this is all I could think of yet. Will try something if and when I get time.

Zburatorul:
@dumbPy, I had a look at the code in #62 and I think I'm missing the logic of how it fits in with the program of moving the prediction to smart_importer and using beancount-import for the UI. Are the extra meta fields indicating a tentative account meant to be inserted by something upstream of the Source defined in #62?

dumbPy:
The `ImporterSource` balances the amounts by adding FIXME account that beancount_import may make a prediction for and the UI allows you to modify it. If you wrap the `ImporterProtocol` subclass in smart_importer's `PredictPostings` hook, it would predict this 2nd posting and  the amount would be anyhow balanced. So now you have got 2 postings, 1 source posting with `source_desc` and `date` metas and one predicted by smart_importer. This predicted posting is passed through without beancount-import's prediction (since it's not a FIXME posting), but I am hoping this/some other PR will make it editable in frontend.

dumbPy:
I suppose the `UNCONFIRMED_ACCOUNT_KEY` is  the missing part in this jigsaw. You may add to `ImporterSource` a func that would add this to any non-source and non-FIXME posting that it may receive from the importer, subject to a bool argument in `__init__`.

Zburatorul:
That makes sense. Thanks for the explanation. I'll try weaving it into the code.

beancount_import/source/paypal.py
```diff
@@ -430,7 +430,7 @@ def _make_import_result(self, txn_id: str, data: Dict[str, Any],
         elif transaction_type_enum.endswith('_SENT') or transaction_type_enum.endswith('_PURCHASE'):
             negate_funding_source_amounts = True
 
-        elif transaction_type_enum.endswith('_RECEIVED'):
+        elif transaction_type_enum.endswith('_RECEIVED') or transaction_type_enum.endswith("EBAY_SALE"):
```
jbms, line 433: Can you split this out into a separate pull request, and add a test?

PR #62, opened 2020-08-02
Add generic importer source

### LINKED ISSUE #18: Generic CSV importer
Hi. I've recently opened an account in a bank that doesn't support OFX export.

I'd like to write a general CSV importer, but I'm having a hard time even starting. Where do I look? I found venmo.py Is probably closest to just CSV, but it's still a bit of a stretch to cut it back to my needs.

CSV structure looks sufficient to uniquely match transaction and have a description and whatnot:

`Account type; Account number; Currency; Date (dd.mm.yyyy); Reference; Description; Credit; Debit`

Reference is somewhat unique and looks like this:

CRD_3629XM - operation with card

A102705190005380 - operation with account

HOLD - you guessed it

Obviously as a lazy person I'd like to just have a working solution in some next release, but steering me in the direction on how to write it myself would be nice too.

### LINKED ISSUE #23: Import from bean-extract / beancount.ingest?

dumbPy:
that imports transactions from `beancount.ingest.importer.ImporterProtocol` subclass importers

performs deduplication w.r.t journal entries and across multiple importable files
tested on my own importer. will be testing more in the following week.

should work as is with #60 
It's my first OSS contribution, so opinions are most welcome @Zburatorul @jbms and others

Closes #18, Closes #23


dumbPy:
I am still facing an issue with manually entered transactions.
```
2020-08-02 * "Headphones"
    Assets:INR:SBI:Saving -2000 INR
    Expenses:Misc
```
and entry from statement imported with this source
```
2020-08-02 * "Some Long ass stupid description"
    Assets:INR:SBI:Saving -2000 INR
        transaction_ref: 34567422445
        source_desc: "Some Long ass stupid description"
    Expenses:FIXME
```

These are matched as expected, but 1st line of both the candidates is `2020-08-02 * "Some Long ass stupid description"`  
I would like to keep `2020-08-02 * "Headphones"` as is from the manually entered entry while merging the postings

This seems to work fine for other sources in your manually_entered example. what am I missing?

jbms:
At the moment there is no indication in the transaction itself whether the narration was manually entered or automatically generated.  I'm not sure how to represent that in a way that doesn't create other problems.

Instead, beancount-import uses a heuristic that if merging an existing transaction in the journal with a new transaction being imported, then the narration from the journal is retained.  However, if both transactions are already in the journal (e.g. one manually entered and one imported with the Expenses:FIXME account left uncorrected), then currently the narration of one will be chosen arbitrarily.

(See matching.py combine_transactions_using_match_set)

Possibly a different heuristic could be used in that case, e.g. preferring the narration from the transaction with the fewest Expenses:FIXME accounts.  Maybe a different heuristic could be used, like checking whether the narration is the same/similar as another metadata field.

beancount_import/source/generic_importer_source.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def prepare(self, journal: 'JournalEditor', results: SourceResults) -> None:
+        results.add_account(self.account)
+        entries = {}
+        for f in self.files:
+            f_entries = self.importer.extract(f)
+            # deduplicate across statements
+            hashed_entries = {}
+            for entry in f_entries:
+                hash_ = self._hash_entry(entry, frozenset(['filename','lineno']))
```
jbms, line 63: Sometimes you can end up with two completely identical transactions, that correspond to two different real transactions, if the datasource doesn't provide a unique transaction id, just a description (e.g. two transactions in the same amount on the same day with the same merchant).  With the type of deduplication you have here, all but one of those transactions will be silently discarded.

The way I've dealt with this in the other data sources is to count for a given source transaction the number of copies in the source data, and expect the same number of copies in the journal.

jbms:
Thanks for this contribution!  Overall this looks good.  Are you able to provide some tests, possibly based on examples in the beancount repository under beancount/ingest/importers?  That would be really helpful in validating that it works an preventing it from breaking in the future.

The tests are relatively easy to create based on the existing framework --- you can refer to e.g. source/mint_test.py

dumbPy:
> Possibly a different heuristic could be used in that case, e.g. preferring the narration from the transaction with the fewest Expenses:FIXME accounts  

I will try this out. and will try to write tests too. Thank you!!!

beancount_import/source/generic_importer_source.py
dumbPy: I actually have a this in my bank statements. Will add hash->count mapping  and use that. Thank you!

dumbPy: Fixed. Now I am maintaining hash->List[Transaction] mapping to cater to this scenario.

dumbPy:
I was using `beancount.ingest.similarity.SimilarityComparator()` for finding transactions in the journal and check if they are cleared.
#### Scenario 1: 
I have 2 very similar transactions in my statement
```
01-01-2020 "by debit card-OTHPG 063441 GOOGLE CLOUD INDIA PVTTHANE-"
    Assets:INR:SBI:Saving -1 INR
    Expense:FIXME

; attention on (063441/063444) difference that will lead to different hash.  
01-01-2020 "by debit card-OTHPG 063444 GOOGLE CLOUD INDIA PVTTHANE-"
    Assets:INR:SBI:Saving -1 INR
    Expense:FIXME
```

but only 1 cleared transaction in journal (say I reconciled midday on 01/01/2020 and then this occurred the next time I tried reconciling).  
The `SimilarityComparator` matches both these transactions to the only cleared transaction in the journal.
This is expected since `SimilarityComparator` will ignore narration that the user might change while reconciling, and everything else is same.
To help with this, now I am matching `source_desc` of the cleared transaction with the imported transactions.

#### Scenario 2:
There are 2 exactly same transactions imported from the statements (i.e., same hash)
```
2020-01-01 * "BULK POSTING- 00000008237 250120 GOOGLE CLOUD INDIA PVT-"
  Assets:INR:SBI:Saving   1.00 INR
  Expenses:FIXME

2020-01-01 * "BULK POSTING- 00000008237 250120 GOOGLE CLOUD INDIA PVT-"
  Assets:INR:SBI:Saving   1.00 INR
  Expenses:FIXME
```
and one cleared transaction in the journal and optionally 1 uncleared(manually entered) transaction.
```
2020-02-26 * "Money returned by Google"
  Assets:INR:SBI:Saving   1.00 INR
    source_desc: "BULK POSTING- 00000008237 250120 GOOGLE CLOUD INDIA PVT-"
  Expenses:Misc          -1.00 INR
```
I am checking the number of cleared transaction that are similar, and have same `source_desc` as imported entry's narration and import the remaining transactions (here 1 since 1 of 2 is already cleared).


Will add these as tests

dumbPy:
Tests added as requested.  
Not implemented `test_invalid` since `InvalidSourceReference` is not used in `ImporterSource`.
I think this is ready to be merged along with #63 
Let me know if you think it needs anything more @jbms

beancount_import/source/generic_importer_source.py
```diff
@@ -0,0 +1,188 @@
...
+            hashed_entries = defaultdict(list)
+            for entry in f_entries:
+                hash_ = self._hash_entry(entry, frozenset(['filename','lineno']))
+                hashed_entries[hash_].append(entry)
+            # deduplicate across statements
+            for hash_ in hashed_entries:
+                # skip the existing entries from other statements. add remaining
+                n = len(entries[hash_])
+                entries[hash_].extend(hashed_entries[hash_][n:])
+
```
jbms, line 78: You could just call get_pending_and_invalid_entries from description_based_source.py rather than using similar_entries_in_journal.  That would avoid having to duplicate similar logic here, and it would also handle invalid references.

beancount_import/source/generic_importer_source.py
```diff
@@ -0,0 +1,188 @@
...
+               and self.is_entry_cleared(entry2) \
+               and entry1.narration == entry2.postings[0].meta['source_desc']
+
+    def _add_description(self, entry: Transaction):
+        if not isinstance(entry, Transaction): return None
+        postings = entry.postings #type: ignore
+        to_mutate = []
+        for i, posting in enumerate(postings):
+            if posting.account != self.account: continue
+            if isinstance(posting.meta, dict): posting.meta["source_desc"] = entry.narration
```
jbms, line 113: I'd recommend also adding 'posting_data' in order to preserve the date from the source data (required in order to reuse get_pending_and_invalid_entries from description_based_source.py).  Otherwise, if this transaction is merged with another one, the date may be lost.  This makes the matching between source entries and existing entries in the journal more robust, since you can do an exact match on the posting date.

beancount_import/source/generic_importer_source.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        if not isinstance(entry, Transaction): return hash_entry(entry)
+        h = hashlib.md5()
+        h.update(hash_entry(entry, exclude_meta=True).encode())
+        for key in entry.meta:
+            if key in exclude_meta_keys: continue
+            h.update(str(entry.meta[key]).encode())
+        return h.hexdigest()
+
+    def is_entry_cleared(self, entry: Transaction) -> bool:
```
jbms, line 150: If you use get_pending_and_invalid_entries from description_based_source.py, you can also just inherit the is_posting_cleared definition from description_based_source.py and not have to define either is_entry_cleared or is_posting_cleared here.

jbms:
Thanks for adding the tests!

This is getting very close --- potentially could be simplified quite a bit by re-using more functionality from description_based_source.py, unless you think there is a reason why something different needs to be done here.

beancount_import/source/generic_importer_source.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        return self._comparator(entry1, entry2) \
+               and self.is_entry_cleared(entry2) \
+               and entry1.narration == entry2.postings[0].meta['source_desc']
+
+    def _add_description(self, entry: Transaction):
+        if not isinstance(entry, Transaction): return None
+        postings = entry.postings #type: ignore
+        to_mutate = []
+        for i, posting in enumerate(postings):
```
jbms, line 111: In your tests the ingester always only generates a single posting, but I imagine in some cases it would generate multiple postings.  In that case, for compatibility with get_pending_and_invalid_entries, the source_desc property should only be added to one of the postings, e.g. the first one.

dumbPy:
I was trying to keep things more close to beancount by using hashing instead of `_get_key_from_entry` and such. But again, I have a habit of over-engineering stuff. All resolved as per your suggestions.

beancount_import/source/generic_importer_source.py
dumbPy: Done by adding break!

beancount_import/source/generic_importer_source.py
dumbPy: Done!

beancount_import/source/generic_importer_source.py
dumbPy: Done! added posting date

beancount_import/source/generic_importer_source.py
dumbPy: Done! I was actually calling `super().is_posting_cleared` after checking `entry.account == self.account`, but I guess that is no more needed since I removed `is_entry_cleared` where I was calling this method for all postings.  
removed.

jbms:
Looks great, thanks!  The only thing remaining is to fix the mypy type annotation errors:

https://travis-ci.com/github/jbms/beancount-import/jobs/369680675#L617

dumbPy:
mypy errors fixed. But the basic_test is failing in python 3.5 but passing in 3.6 and 3.7  
I wonder what went wrong :cry:

jbms:
It looks like the issue is that the order of the entries happens to differ in Python 3.5 --- in Python 3.5, iterating over a dict returns elements in a random order.  In python 3.6 and later, it is guaranteed to be in insertion order.  To ensure a deterministic result, you could use OrderedDict instead of defaultdict.  Unfortunately you lose the convenience of defaultdict then, but you can just use the setdefault method which is almost as convenience.

dumbPy:
Fixed!!

jbms:
Thanks!

One other thing would be to add this source to the list of sources in the README.md file, and to expand the comment at the top of the file to include for detailed instructions on how to use it, like the other data sources have, but that can be a separate PR.

PR #63, opened 2020-08-06
Journal entry filename should have bean in name

dumbPy:
Continuing conversation from [here](https://github.com/jbms/beancount-import/pull/62#issuecomment-667668211)

`beancount.ingest.importer.ImporterProtocol` based importers add `filename` to meta of each transaction.  
Hence they are being classified as journal entry too.  
This causes the beancount-import to pick Imported entry's narration over manually entered transaction's narration.  
Since journal has `.bean/.beancount` extension, that can be an added filter. Statement names almost never contain the word `bean`.  

Tested.

### DIFF
--- a/beancount_import/matching.py
+++ b/beancount_import/matching.py
@@ -120,6 +120,7 @@ set.  The resulting sum must equal 0 within a small tolerance.
 
 """
 
+import os
 import datetime
 import collections
 import itertools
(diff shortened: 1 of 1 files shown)


jbms:
I think it might be better to just strip off the filename metadata when passing the entries to beancount-import, since there isn't really a standard filename extension for beancount.

dumbPy:
Handled in ImporterSource #62

PR #65, opened 2020-08-08
 Support Importing Balance Directive for ImporterSource

dumbPy:
`beancount.ingest.importers.csv.Importer` also returns Balance directive when available.  
add support for this in `ImporterSource`
Tested.


jbms:
Thanks, and I apologize for the delay in responding.

PR #72, opened 2020-10-18
Amazon enhancements (caching, date filter, groceries support)

mbafford:
Several enhancements to the Amazon processing code:

* Caching of HTML parse results as pickle files
* Configuration option to skip invoices before a certain date (which can only act after parsing the HTML)
* Support for grocery orders through Amazon Fresh, Whole Foods, or Prime Now which contain tips (with no currency specified in the HTML) and weighed items with slightly different quantity and price formats

With a hundreds of Amazon files, and a lot of them being excluded due to the date filter (and thus can't be skipped due to existing in my journal), the caching makes a huge difference in startup/processing time.

Hopefully the parsing changes didn't break anything for non-US customers. I haven't encountered any issues, and my changes pass the existing tests, but I'm not sure how much coverage those tests provide.

### DIFF
--- a/beancount_import/amount_parsing.py
+++ b/beancount_import/amount_parsing.py
@@ -25,18 +25,20 @@ def parse_number(x):
     sign, number_str = parse_possible_negative(x)
     return sign * D(number_str)
 
-def parse_amount(x):
+def parse_amount(x, assumed_currency=None):
     """Parses a number and currency."""
     if not x:
         return None
     sign, amount_str = parse_possible_negative(x)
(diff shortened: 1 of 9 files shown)


mbafford:
Updated to address mypy concerns. Still fails build on Windows, but that seems to be impacting other branches and not directly related to my changes.

jbms:
Thanks

arnold-c:
Just a quick note for anyone who may be having an issue implementing the Amazon caching - you have to be very careful about how you specify the path i.e. `pickle_dir=os.path.join(cache_dir, 'amazon/')` must end in `/` otherwise it will fail.

PR #74, opened 2020-12-05
Fix generic importer tests that implicitly depend on libmagic

carljm:
The generic importer source takes a directory, looks at all files in
that directory recursively, and then filters down to the files that are
of the right type for its importer. The tests for it point it to a
directory that contains not only the test source CSV file, but also the
test verification beancount and JSON files.

I haven't verified, but I'm guessing this worked for the original author
because they had `python-magic` and `libmagic` installed. But without
those installed, beancount's file-identification fails (ironically) on
the `.beancount` files, causing these tests to fail.

Since `libmagic` is supposed to be an optional dependency of beancount,
and it can't be listed as an automatic test dependency (since it's not a
Python library), the tests should not require it in order to pass.

This behavior of the generic importer (silently excluding files of a
type it doesn't know how to handle) seems a bit too magical and implicit
for my tastes, but since I didn't author and don't use this importer I'm
not changing its behavior. Instead I just rearranged the test data
directory so the source CSV file is in its own subdirectory, eliminating
the problem of making the importer identify the various other test
support files.

With this PR all tests are green again, even without `libmagic`
installed.


carljm:
cc @dumbPy, author of the tests in question here

jbms:
Thanks!

dumbPy:
Hi @carljm 
Thank you for this fix. You are right, I use anaconda so I didn't actually check the dependencies.
> This behavior of the generic importer (silently excluding files of a
> type it doesn't know how to handle) seems a bit too magical and implicit
> for my tastes, but since I didn't author and don't use this importer I'm
> not changing its behavior. 

I am actually using `beancount.ingest.importers.csv.Importer` in the test and the file filtering is actually baked into beancount and with that, this implicit behaviour, I suppose. Nonetheless, Thank you for the fix.

PR #71, opened 2020-10-18
Fix pytest mimetype issues with generic_importer_source_test.py

mbafford:
Without catching these exceptions, I get these errors when running `pytest`:

```
FAILED beancount_import/source/generic_importer_source_test.py::test_source[test_basic] - ValueError: Could not identify the type of file 'testdata/source/generic_importer/test_training_examples/journal.beancount'
FAILED beancount_import/source/generic_importer_source_test.py::test_source[test_invalid] - ValueError: Could not identify the type of file 'testdata/source/generic_importer/test_training_examples/journal.beancount'
FAILED beancount_import/source/generic_importer_source_test.py::test_source[test_training_examples] - ValueError: Could not identify the type of file 'testdata/source/generic_importer/test_training_examples/journal.beancount'
```

The error is coming from:

https://github.com/beancount/beancount/blob/0aed556e5c838e07ef0566af8199c9ae28e9a639/beancount/utils/file_type.py#L68

It's not clear to me why this isn't cropping up for other people. I would be happy to fix in a different way, I just couldn't test my other PR properly fully without fixing this.

This is on macOS 10.15.6 Catalina with Python 3.7.4.

With this combination of dependencies:
```
beancount==2.3.2
python-magic==0.4.18
atomicwrites==1.4.0
attrs==20.2.0
beautifulsoup4==4.9.3
bottle==0.12.18
cachetools==4.1.1
certifi==2020.6.20
chardet==3.0.4
click==7.1.2
google-api-core==1.22.4
google-api-python-client==1.12.3
google-auth==1.22.1
google-auth-httplib2==0.0.4
googleapis-common-protos==1.52.0
httplib2==0.18.1
idna==2.10
importlib-metadata==2.0.0
iniconfig==1.1.1
joblib==0.17.0
jsonschema==3.2.0
lxml==4.6.0
nltk==3.5
numpy==1.19.2
packaging==20.4
pathtools==0.1.2
pluggy==0.13.1
ply==3.11
protobuf==3.13.0
py==1.9.0
pyasn1==0.4.8
pyasn1-modules==0.2.8
pyparsing==2.4.7
pyrsistent==0.17.3
pytest==6.1.1
python-dateutil==2.8.1
pytz==2020.1
regex==2020.10.15
requests==2.24.0
rsa==4.6
scikit-learn==0.23.2
scipy==1.5.3
six==1.15.0
soupsieve==2.0.1
threadpoolctl==2.1.0
toml==0.10.1
tornado==6.0.4
[...]

PR #75, opened 2020-12-06
[amazon] fix pretax adjustments not in shipments

carljm:
Judging by included testdata invoices, it seems that Amazon invoices
used to include pretax adjustments (e.g. S&H, Subscribe & Save savings,
etc) both in each individual shipment summary and in the final payments
block. But modern Amazon invoices only include these once in the
payments block, not in the shipments info.

Before this diff, the modern-style invoices would result in a
transaction with an error about grand total amount mismatch, because the
pretax adjustments in the payments block were not included in the
calculation of the expected grand total.

This commit fixes that without breaking parsing of the older style of
invoices (all existing tests pass without changes). The logic is that we
track the total amount of payment-block pretax adjustments. If we find
adjustments within shipments, that indicates an older style invoice, and
we assert that those total the same amount as the pretax adjustments in
the payments block, then we subtract the total pretax adjustments for
shipments from the tracked total from payments. For a well-formed
old-style invoice, this should leave us with zero in
`payments_pretax_adjustment_total`, so there ends up being no change in
the expected grand total.

### DIFF
--- a/beancount_import/source/amazon_invoice.py
+++ b/beancount_import/source/amazon_invoice.py
@@ -358,7 +358,7 @@ def parse_regular_order_invoice(path: str) -> Order:
     output_fields = dict()
     output_fields['pretax_adjustments'] = get_adjustments_in_table(
         payment_table, pretax_adjustment_fields_pattern)
-    amount = reduce_amounts(
+    payments_pretax_adjustment_total = reduce_amounts(
         a.amount for a in output_fields['pretax_adjustments'])
     payment_adjustments = collections.OrderedDict()  # type: Dict[str, Amount]
     if any(s.pretax_adjustments for s in shipments):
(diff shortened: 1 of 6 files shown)


mbafford:
Confirmed this is an issue. I ran into the same thing (my fix for that is bundled in the other changes #71 ).

carljm:
> (my fix for that is bundled in the other changes #71 ).

Doesn't look like it's in #71, I'm guessing you meant #72?

mbafford:
> Doesn't look like it's in #71, I'm guessing you meant #72?

Whoops. Yep, #72 - specifically https://github.com/jbms/beancount-import/pull/72/commits/c0f71c4877b47a9ba07cd713faccfa0c2322fcd9

carljm:
Nice. Our fixes look pretty much identical in principle, just implementation details differ.

jbms:
Thanks, I merged #72 instead since per the discussion it provides a similar fix but also includes some other functionality.

I apologize for the delay in merging #72, which led to you duplicating the work there.

Thanks for all your contributions lately, though!

PR #76, opened 2020-12-06
[matching] allow grouping postings of differing sign

carljm:
My use case for allowing this is importing transactions from Amazon,
which can include e.g. Subscribe & Save or other discounts along with
the cost of the item and tax, all in the same Expenses:FIXME:A account,
balancing a credit card posting for the net cost. In order for this to
match the corresponding transaction from my credit card account, we must
allow grouping the item cost, tax, and discounts into a single
MatchablePosting whose weight is the net cost, which can match the
unknown-expense leg of the credit card transaction. Currently such a
grouped MatchablePosting is disallowed by the "same sign" rule.

I verified that this change works correctly for matching transactions of
the above form. It doesn't break any of the existing matching tests.

I'm curious if there are use cases in which the "same sign" rule is
critical to avoid wrong matches? It seems like the net effect of a set
of postings all to the same account is something that should in general
be considered for matching, even if the postings are not all of the same
sign.

Feedback / counter-examples welcome!

### DIFF
--- a/beancount_import/matching.py
+++ b/beancount_import/matching.py
@@ -704,7 +704,7 @@ def get_aggregate_posting_candidates(
     5. Subsets may not sum to zero, or contain any sub-subsets that sum to zero.
 
     6. To limit the computational cost, subsets are limited to at most 4
-       elements, except that all maximal subsets are also returned.
+       elements, except that all same-sign maximal subsets are also returned.
 
     The returned subsets are not, in general, disjoint.
 
@@ -730,15 +730,26 @@ def get_aggregate_posting_candidates(
     def posting_set_id(postings):
         return tuple(id(x) for x in postings)
 
-    def add_subset(account, currency, subset):
+    def partition(predicate, postings):
+        t = []
+        f = []
+        for p in postings:
+            if predicate(p):
+                t.append(p)
+            else:
+                f.append(p)
+        return t, f
+
+    def add_subset(account, currency, subset, check_zero=True):
         total = sum(x.units.number for x in subset)
-        if total == ZERO:
-            sum_to_zero.add(posting_set_id(subset))
-            return
-        for subsubset_size in range(2, len(subset)):
-            for subsubset in itertools.combinations(subset, subsubset_size):
-                if posting_set_id(subsubset) in sum_to_zero:
-                    return
+        if check_zero:
+            if total == ZERO:
+                sum_to_zero.add(posting_set_id(subset))
+                return
+            for subsubset_size in range(2, len(subset)):
+                for subsubset in itertools.combinations(subset, subsubset_size):
+                    if posting_set_id(subsubset) in sum_to_zero:
+                        return
         aggregate_posting = Posting(
             account=account,
             units=Amount(currency=currency, number=total),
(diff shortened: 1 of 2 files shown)


jbms:
It has been a few years since I've thought about this much, but I think the reason I added the same sign rule was to prevent groups from containing additional unrelated postings that just cancel each other out.

For example:

```
2020-12-05 * "Amazon transaction"
  Expenses:FIXME  20 USD ; item 1
  Expenses:FIXME  5 USD ; shipping 1
  Expenses:FIXME  30 USD ; item 2
  Expenses:FIXME  7 USD ; shipping 2
  Expenses:FIXME -25 USD ; payment for item 1
  Expenses:FIXME -37 USD ; payment for item 2
```

Without the sign rule, if we have a -37 USD posting from the bank to match, it would match both:

  Expenses:FIXME -37 USD ; payment for item 2

But also the group:

```
  Expenses:FIXME  20 USD ; item 1
  Expenses:FIXME  5 USD ; shipping 1
  Expenses:FIXME -25 USD ; payment for item 1
  Expenses:FIXME -37 USD ; payment for item 2
```

Note: I haven't taken the time to think through this example too carefully, though.  In general it seems that there is also just a much larger chance of spurious matches if mixed signs are allowed.

One option could be to define a metadata property for tagging postings to allow mixed sign groups, and then the amazon source could use that.

carljm:
Makes sense, thanks for the thoughtful response. So in practice this isn't a problem for the Amazon importer, since it generates costs/discounts using accounts of the form `Expenses:FIXME:X` where `X` is `A`, `B`, `C` etc for different shipments, whereas it generates the payments using just `Expenses:FIXME`. So the same-account restriction is sufficient in this case to avoid spurious matches. But it does seem possible that the scenario you outlined could occur in an importer that doesn't group FIXME accounts in this way.

The conceptually simple (but perhaps not simple to implement) generic fix I can imagine would be to disallow any group containing any subgroup that sums to zero. Are you open to this option if I'm able to implement it reasonably?

Failing that, restricting this to the Amazon importer via a metadata property also seems reasonable to limit the blast radius.

(By the way, while I've got you, thanks so much for beancount-import and finance-dl! They are absolutely crucial to making my beancount workflow manageable in limited available time.)

jbms:
I think your idea of disallowing any subgroup to sum to 0 is a reasonable constraint.  The only issue is that it is unavoidably exponential cost in the size of the set to check if a subset sums to 0 (https://en.wikipedia.org/wiki/Subset_sum_problem).  Fortunately, subsets are currently limited to size 4, so exponential time wouldn't be be too bad.  You could also skip this check if the group happens to be single-sign.

The other thing is that currently in addition to subsets of size <= 4, a candidate group is also generated for all postings for a given (account, currency, sign).  That was intended to partially mitigate the restriction of subset size <= 4, as I figured that might be a common case.  It would probably be good to preserve that behavior even after you otherwise remove the sign restriction.

carljm:
Ok, pushed an implementation of the approach we discussed, with tests. Let me know what you think!

jbms:
Thanks!

PR #78, opened 2020-12-18
Add support for EBAY_SALE and INVOICE_PAID in PayPal importer

carljm:
Added a test transaction for each; tests and mypy pass.

INVOICE_PAID is trivial, it just works with no adjustment. EBAY_SALE requires a few adjustments but isn't too bad.


jbms:
Thanks!

PR #80, opened 2020-12-18
More Schwab action types (Margin interest, and some option related)

Zburatorul:
Margin interest looks just like a fee. Options are trickier. 

One issue is that the "symbol" reported by Schwab (`SPY 06/19/2020 260.00 P`) is not a valid beancount symbol. I chose to strip any non-alphanumeric characters, which results in long and hard to read symbol names.

I chose to reuse the Buy and Sell classes originally created for stocks. This is mostly fine, but there are a couple of situations I haven't handled yet. "Sell to Open" events should not have an Income related posting, because no gain/loss is there to be recorded; "Buy to Close" should have such a posting.

Finally, options are sold in lots of one hundred, while their price is reported for one unit. This means that the option quantity bought/sold that Schwab reports times the price, will not match the market value. Either the quantity needs to be multiplied by 100, or the price. This is not handled yet.


beancount_import/source/schwab_csv.py
```diff
@@ -683,6 +703,7 @@ def process_positions(
 TAXES_ACCOUNT_KEY = "taxes_account"
 DATE_FORMAT = "%m/%d/%Y"
 TITLE_RE = re.compile(r'"Transactions  for account (?P<account>.+) as of (?P<when>.+)"')
+SYMBOL_RE = re.compile(r'[^\d\w]')
```
carljm, line 706: Can we name this `STRIP_FROM_SYMBOL_RE` to clarify that it is not a regex to match a valid symbol name, but rather a regex to match invalid characters?

beancount_import/source/schwab_csv_test.py
```diff
@@ -21,8 +21,8 @@ def test_source(name: str):
         example_dir=example_dir,
         source_spec={
             'module': 'beancount_import.source.schwab_csv',
-            "transaction_csv_filenames": glob.glob(f"{example_dir}/transactions/*.CSV"),
-            "position_csv_filenames": glob.glob(f"{example_dir}/positions/*.CSV"),
+            "transaction_csv_filenames": sorted(glob.glob(f"{example_dir}/transactions/*.CSV")),
```
carljm, line 24: You can rebase your branch on latest master with #79 merged to eliminate this change from this PR.

testdata/source/schwab_csv/test_basic/accounts.txt
```diff
@@ -5,3 +5,6 @@ Assets:Schwab:Intelligent-4321
 Assets:Schwab:Intelligent-4321:Cash
 Assets:Schwab:Intelligent-4321:FNDA
 Assets:Schwab:Intelligent-4321:SCHA
+Assets:Schwab:Intelligent-4321:SMAL040120203200P
```
carljm, line 8: I would see some value in converting option symbol names in the same way that the Schwab OFX export used to do. Do these match previous transactions in your journal that were imported via Schwab OFX?

carljm (commented):
Code looks generally good to me so far! Left a few comments. Regarding the symbol names, I left a comment that it might be good to match whatever Schwab OFX import did previously. Reusing Buy/Sell looks fine so far, we can see what it looks like after you add the posting changes.

testdata/source/schwab_csv/test_basic/accounts.txt
Zburatorul: I think the OFX importer does no conversion. I always had to go edit the symbol names by hand.
Perhaps a small improvement over my solution is to replace spaces with dashes, but drop the slashes.

beancount_import/source/schwab_csv.py
Zburatorul: Done.

beancount_import/source/schwab_csv_test.py
Zburatorul: Done.

Zburatorul: I did rebase just now before pushing the latest changes. Not sure why this is still showing up.

beancount_import/source/schwab_csv.py
```diff
@@ -343,7 +350,6 @@ def get_meta(self) -> Meta:
     def get_narration_prefix(self) -> str:
         raise NotImplementedError()
 
-
```
carljm, line 346: Nit: two lines between top level classes is typical and intentional formatting, please restore this blank line.

beancount_import/source/schwab_csv.py
```diff
@@ -371,7 +377,14 @@ def get_other_account(self) -> str:
     def get_narration_prefix(self) -> str:
         return "INVBANKTRAN"
```
carljm, line 373: And add blank line here.

beancount_import/source/schwab_csv.py
```diff
@@ -371,7 +377,14 @@ def get_other_account(self) -> str:
...
 
+@dataclass(frozen=True)
+class MarginInterest(TransactionEntry):
+    def get_sub_account(self) -> Optional[str]:
+        return "Cash"
 
+    def get_narration_prefix(self) -> str:
+        return "MARGININTEREST"
+
```
carljm, line 387: And here.

beancount_import/source/schwab_csv_test.py
carljm: Hmm, doesn't seem to have pushed with the rebase, since I still see these changes here.

testdata/source/schwab_csv/test_basic/accounts.txt
carljm: Oh in that case I think whatever you want to do is fine. Dashes might help readability a bit if they are allowed in beancount.

beancount_import/source/schwab_csv_test.py
Zburatorul: Solved now.

testdata/source/schwab_csv/test_basic/accounts.txt
Zburatorul: Turns out no dashes in Beancount commodities, so simplest to strip everything.

PR #81, opened 2020-12-22
Schwab CSV: parse per-account positions CSV too.

carljm:
Originally I was manually downloading Schwab CSV, so it was simpler to download the "All-Accounts" positions CSV, so I wrote the importer to expect that. In writing a finance-dl downloader, I realized that the best data storage format would keep each account in its own directory, and it was simpler to download both transactions and positions per-account.

So this PR updates the Schwab CSV source so it can handle either type of positions CSV. Added test file for the single-account format, and the existing test file already covers All-Accounts variant.

Tox passes clean.


carljm:
Merged in master and resolved the conflict with the other PR; tox is green again.

PR #82, opened 2020-12-22
Add UNCONFORMED_ACCOUNT_KEY to imported entries

dumbPy:
If a posting was predicted by smart_importer or was entered by hand, it won't be predicted again but should be editable like the FIXME or predicted entry by beancount-import.

This was suggested and discussed in #60 and will start working once #60 is completed and merged.


jbms:
Please change variable name to UNCONFIRMED

Zburatorul:
@dumbPy, glad you went ahead with this. How do I set up the minimal example to test that #62 works?

dumbPy:
@Zburatorul use below code as `generic_importer_source_test.py`. It will add dummy postings to all the imported transactions just like the smart_importer. the basic_test fails with it, but the other two tests pass. 

You can also use the `DummyPostingPredictor` from below with any other importer and statements you have, and it will add the balanced Dummy posting. #82 adds `{"unconfirmed_account":""}` to meta of this Dummy Posting. #60 should make this posting editable from frontend.

```python
from decimal import Decimal
import os

import pytest

from .source_test import check_source_example
from beancount.ingest.importers.csv import Importer as CSVImporter, Col
from beancount.core.data import Transaction, Posting, Amount

testdata_dir = os.path.realpath(
    os.path.join(
        os.path.dirname(__file__), '..', '..', 'testdata', 'source', 'generic_importer'))

testdata_csv = os.path.join(testdata_dir, "csv")

examples = [
    'test_basic',
    'test_invalid',
    'test_training_examples'
]

importer = CSVImporter({Col.DATE: 'Date',
                        Col.NARRATION1: 'Description',
                        Col.AMOUNT: 'Amount',
                        },
                       'Assets:Bank',
                       'USD',
                       '"Date","Description","Amount"',
                       )

class DummyPostingPredictor:
    DUMMY_ACCOUNT = "Assets:Dummy"

    def __init__(self, importer):
        self.importer = importer
        # move extract to _extract
        self.importer._extract = self.importer.extract
        self.importer.extract = self.extract_wrapper
    
    def extract_wrapper(self, f, existing_entries):
        entries = self.importer._extract(f, existing_entries)
        for entry in entries:
            if isinstance(entry, Transaction):
                p = entry.postings[0]
                entry.postings.append(
                    Posting(
                        account=self.DUMMY_ACCOUNT,
                        units=Amount(currency=p.units.currency, number=-1*p.units.number),
                        cost=None,
                        price=None,
                        flag=None,
                        meta={},
                    )
                )
        return entries

# wrap the importer with the dummy posting predictor.
# Will predict 2nd dummy postings for the importer entries
DummyPostingPredictor(importer)

@pytest.mark.parametrize('name', examples)
def test_source(name: str):
    check_source_example(
        example_dir=os.path.join(testdata_dir, name),
        source_spec={
            'module': 'beancount_import.source.generic_importer_source',
            'directory': testdata_csv,
            'account': 'Assets:Bank',
            'importer': importer,
        },
        replacements=[(testdata_dir, '<testdata>')])

```

beancount_import/source/generic_importer_source.py
```diff
@@ -84,24 +85,20 @@ def prepare(self, journal: 'JournalEditor', results: SourceResults) -> None:
...
             else:
-                to_mutate.append(i)
-                break
-        for i in to_mutate:
-            p = postings.pop(i)
-            p = Posting(p.account, p.units, p.cost, p.price, p.flag,
-                        {"source_desc":entry.narration, "date": entry.date})
-            postings.insert(i, p)
+                # add unconfirmed account key
+                posting.meta[UNCONFIRMED] = ""
```
jbms, line 101: Perhaps use `True` rather than `""` since this seems to be a boolean field?

jbms:
Can you please add a unit test?

beancount_import/source/generic_importer_source.py
jbms, line 101: It looks like currently all accounts other than `self.account` are considered to be editable.  Should there be a way for the importer to indicate a non-editable account?

testdata/source/generic_importer/test_dummy_predictor/import_results.beancount
```diff
@@ -0,0 +1,65 @@
+;; date: 2020-01-01
+;; info: {"filename": "<testdata>/csv/generic_statement.csv", "line": 1, "type": "text/csv"}
+
+; features: []
+2020-01-01 * "by debit card-OTHPG 063441 GOOGLE CLOUD INDIA PVTTHANE-"
+  Assets:Bank   -1 USD
+    date: 2020-01-01
+    source_desc: "by debit card-OTHPG 063441 GOOGLE CLOUD INDIA PVTTHANE-"
+  Assets:Dummy   1 USD
+    unconfirmed_account: ""
```
jbms, line 10: It looks like the test output needs to be updated for the change to `True`.

dumbPy: Done

beancount_import/source/generic_importer_source.py
dumbPy: Wouldn't it be hard to differentiate without smart_importer adding some meta to the postings? I can check for `"confirmed" : True` key in a posting meta before adding the unconfirmed key. So if an importer or `smart_importer` adds 2nd/nth posting abd wants to make a posting non-editable, they can add the confirmed tag, but there wouldn't be a way to enforce it.

Zburatorul: I personally don't yet see the utility of this. Even if a posting is marked as editable by the generic_importer, one can safely ignore it in the beancount-import interface. I admit it is slightly more cognitively taxing to have to review an extra candidate posting.

jbms: Yes, the idea would be to reduce cognitive load.  Accounts that are known from the structure of the data itself (e.g. investment transactions where some legs are known to be interest, dividends, capital gains, fees, etc.) don't need to be inspected manually, while any "guessed" accounts may well be wrong and need to be manually checked somewhat carefully, which takes longer and more mental effort.

I suppose currently smart_importer does not provide any indication of which accounts were "guessed" vs determined from the structure of the data?  Could it be made to do so, i.e. add unconfirmed_account: True?  Then it would not be necessary for beancount-import to add it.

It also seems like it would be better if the unconfirmed_account metadata field were stripped out before actually writing the transaction to the journal, and it would be kind of confusing to leave it after the account has actually been confirmed.  The exception would be the FIXME action --- that should leave unconfirmed_account in place.  Then the presence of unconfirmed_account in the journal should be treated like an Expenses:FIXME account, in that it should be presented to the user as a pending transaction.

In general it might be easier if one of you, @Zburatorul or @dumbPy, could combine this pull request with #60 as a single pull request (and ideally add an example and test of using this with smart_importer if that is the ultimate goal), so that it is easier to test that it all actually works together in a desirable way.

Zburatorul: I see the logic now. Shall we for now move forward assuming that smart_importer will insert the unconfirmed_account key for any uncertain guesses? The absence will be assumed to mean certainty.

I implemented the logic you suggest of removing the "unconfirmed_account" key from postings with accounts other than FIXME upon acceptance in the UI.

@dumbPy, since my changes are smaller and simpler than yours I suggest you cherry-pick them onto your branch and modify at will.

dumbPy:
@jbms @Zburatorul  Hey guys, this feels ready now.
Sorry I dropped out of the loop 4 years ago :p
Cherry-picked from #60 
Now, if you wrap the importers in smart_importer's predictors,
```
for importer in CONFIG:
    apply_hooks(importer, [PredictPostings(), PredictPayees()])
```
1. The imported transactions are added posting by `PredictPostings()`
2. [generic_importer_source.py](https://github.com/jbms/beancount-import/pull/82/files#diff-127ae9abdac941d936f3929fc61ef6a9fa0abb1fd038905af4a28165892e5df2R103) adds `"unconfirmed_account":True` metadata to the posting.
3. [reconcile.py](https://github.com/jbms/beancount-import/pull/82/files#diff-d2af8c1d9348e25b9f515f127a69f1600b861955b86e0d689fd6d17e5d0d46f5R785) checks for this `"unconfirmed_account":True`, pops the posting, and adds FIXME posting in it's place and uses the corresponding account as dummy prediction.
4. This dummy prediction is now passed forward as if it was predicted by the beancount-import's predictor allowing editing in the frontend.

dumbPy:
@Zburatorul  I have been using this too for a few months and works perfectly fine as expected.

PR #83, opened 2020-12-22
Use empty posting while balancing imported entry

dumbPy:
If an empty posting i.e., one without an amount, is entered by hand,
it should be used to balance the transaction instead of a FIXME posting.

### Example
```
2020-08-26 * "Loan to Asim"
    Assets:INR:HSBC:Saving -34000 INR
    Assets:Receivables:Asim
```
when this transaction is imported currently, it gets balanced as
```
2020-08-26 * "Loan to Asim"
    Assets:INR:HSBC:Saving -34000 INR
    Assets:Receivables:Asim 
    Assets:FIXME            34000 INR
``` 
but should be, after this PR
```
2020-08-26 * "Loan to Asim"
    Assets:INR:HSBC:Saving -34000 INR
    Assets:Receivables:Asim 34000 INR
```

### DIFF
--- a/beancount_import/source/generic_importer_source.py
+++ b/beancount_import/source/generic_importer_source.py
@@ -139,21 +139,40 @@ def get_info(raw_entry: Directive) -> dict:
         line=raw_entry.meta['lineno'],
     )
 
+def _none_units_posting_index(txn:Transaction):
+    """Index of posting with None units"""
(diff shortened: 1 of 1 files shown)


jbms:
Thanks, this looks good, but can you please add a unit test for this?

Issue #84, opened 2020-12-22
generic_importer_source examples fail on relative path


Zburatorul:
Freshly checked out master. Same issue for the manually_entered examples.
@dumbPy 
```
eugeniu@home:~/beancount-import/examples/fresh$ python3 run.py 
Listening at http://127.0.0.1:8101
../data/importers/creditcard.csv
Traceback (most recent call last):
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/webserver.py", line 502, in _handle_reconciler_loaded
    loaded_reconciler = loaded_future.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/thread_helpers.py", line 13, in wrapper
    f.set_result(fn(*args, **kwargs))
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/reconcile.py", line 382, in __init__
    self._load_sources()
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/reconcile.py", line 435, in _load_sources
    sources = self.sources = [
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/reconcile.py", line 436, in <listcomp>
    load_source(spec, log_status=self.reconciler.log_status)
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/source/__init__.py", line 319, in load_source
    return m.load(source_spec, log_status=log_status)  # type: ignore
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/source/generic_importer_source.py", line 160, in load
    return ImporterSource(log_status=log_status, **spec)
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/source/generic_importer_source.py", line 45, in __init__
    files = [get_file(f) for f in
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount_import/source/generic_importer_source.py", line 45, in <listcomp>
    files = [get_file(f) for f in
  File "/home/eugeniu/.local/lib/python3.8/site-packages/beancount/ingest/cache.py", line 136, in get_file
    raise e
[...]

PR #85, opened 2020-12-22
Use absolute path to data file in generic_importer_source

Zburatorul:
Fixes issue #84.


jbms:
Maybe instead generic_importer_source should call abspath?

Issue #88, opened 2020-12-31
Some work left to do on the Schwab importer


Zburatorul:
I wanted to collect a few loose ends here:

* Add commodity directives for every new commodity that appears in transactions.
* Handle bonds, and US Treasuries specifically. Bond symbols are CUSIPs, which are not valid Beancount symbols. Also accounting seems a bit funkier.
* Allow users to specify if commodity transactions should be in subaccount, or stay in top-level account like OFX import does it.
* Option expiration.
* Option sold in lots of 100, even though price reported for 1.
* Add/skip Income account posting for capital gains (no need when selling short, but needed when buying to close). 

@carljm 
Comments and suggestions are welcome.

carljm:
Sounds reasonable to me; I don't have experience with most of those transaction types, so I can review the code but won't have much input on correctness of the handling. Thanks for filling in the gaps!

bayesianmind:
I have a PR in the works to add Schwab Bank Checking support.

zhou13:
It seems that schwab_csv importer does not support `Reinvest Shares` and `Reinvest Dividend`.  I am getting errors like 

````
ValueError: 'Reinvest Shares' is not a valid SchwabAction
````

The csv looks like (date and amount are modified)
````
"Date","Action","Symbol","Description","Quantity","Price","Fees & Comm","Amount",
"00/00/0000","Reinvest Dividend","SWVXX","SCHWAB VALUE ADVANTAGE MONEY FUND","","","","$0.1",
"00/00/0000","Reinvest Shares","SWVXX","SCHWAB VALUE ADVANTAGE MONEY FUND","0.1","$1.00","","-$0.1",
Transactions Total,"","","","","","",$0.00
````

bayesianmind:
#91 Adds support for Reinvest Shares and Reinvest dividend among other new action types.
@Zburatorul Do you mind taking a look at that PR and see what you think of it?

Zburatorul:
> #91 Adds support for Reinvest Shares and Reinvest dividend among other new action types.
> @Zburatorul Do you mind taking a look at that PR and see what you think of it?
I had a look and I don't have any comments.

Zburatorul:
A few more actions types that need implementation:

1. `Short Term Cap Gain`
2. `Long Term Cap Gain`
3. `Stock Merger`

zhou13:
[...]

PR #89, opened 2020-12-31
Schwab importer option lots and capital gains postings

Zburatorul:
Solved two issues mentioned in #88.

Schwab reports buying 1 option for $5, but amount is $500 because options are actually sold in batches of 100. I chose to increase the quantity 100X rather than price because price may be obtained through other sources in the future.

The second issue is that selling short does not require an Income posting, while buying to cover does.


carljm (approved):
It would be good to add sell-to-open and buy-to-close transactions to the test examples, so we can verify the effect of the conditional cap-gains posting in those cases. Otherwise this looks good to me!

Zburatorul:
> It would be good to add sell-to-open and buy-to-close transactions to the test examples, so we can verify the effect of the conditional cap-gains posting in those cases. Otherwise this looks good to me!

We do. Lines 20-21 of Intelligent_Transactions.

carljm:
Oh, indeed! And now I see also that the cap-gains posting is removed from the sell short transaction and added to the buy-cover transaction, as should be expected. Not sure how I missed that in last review. Looks great to me!

PR #90, opened 2021-01-08
Use WSL for pdftotext on Windows

bayesianmind:
pdftotext is not distributed for Windows and I was unable to get
it working as a native version even after substantial effort.
In contrast it's quite easy to get via WSL (Windows Subsystem for Linux)
where you can install it in any of the available Linux distributions.

### DIFF
--- a/beancount_import/source/ultipro_google_statement.py
+++ b/beancount_import/source/ultipro_google_statement.py
@@ -1,5 +1,5 @@
 """Parses a Google employee PDF pay statement from Ultipro."""
-
+import os
 from typing import NamedTuple, Dict, Any, List, Optional, Tuple, Union, Callable, Match
 import datetime
 import collections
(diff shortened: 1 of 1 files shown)


jbms:
Using WSL seems reasonable --- I suppose cygwin could also be used maybe?

I wonder if instead it would make sense to just create a pdftotext script that you add to your PATH that forwards to wsl, and that could be included in beancount-import or documented somehow as a recommended method.  That way beancount-import itself can be agnostic to how pdfottext is invoked.  Another option could be to create an environment variable that specifies the pdftotext executable name --- which would allow you to do the same thing without the need to add it to your PATH.

bayesianmind:
The environment variable approach is nice, I like that idea.
I am not sure about the viability of the script approach - I don't know
that Windows (outside WSL) has shebang or a way to easily make a forwarding
script like that that would execute correctly without the right file
extension. You can obviously build a custom binary to do it but that's a
pain.
I'll patch this to take the envvar approach.

PR #91, opened 2021-01-11
Add Schwab bank support to schwab_csv importer

bayesianmind:
I have quite a lot of activity / business with Schwab including a checking account so I brought in support for everything I could find in my own Schwab CSV files. In the spirit of #88 .

Other changes in the schwab_csv source: (apologies for not splitting these out, if that's a bother I can do the extra book keeping to get them into independent pull reqs).
- Added support for 7 new BrokerageActions.
- Added support for fractional shares, which Schwab added last year. You can get these in your account via dividend re-investment or the fractional share purchase feature. The source previously assume int for share quantities.
- Make schwab_csv post cleared status for Schwab accounts. This lets uncleared transactions properly appear in the import webui.
- Make the currency configurable just in case instead of hardcoding USD.
- Added a dedicated interest type that we now use consistently for Checking account interest and "Bank interest" paid on brokerage account cash balances.
- Tweaked some of the key regexps to work on the files I am getting from Schwab in addition to the testfiles.

Known caveats:
- The banking CSVs indicate they may contain pending transactions in a special way. I didn't have any in my file to add support for. Likely the importer will import them the same whether they are pending or not. This could result in some extraneous imports for pending auth charges that would normally drop off. If someone donates a CSV with the format it would be easy to add support for it.
- Does not support corporate equity awards accounts - for Google's at least it is quite different from the Brokerage format and I had written my own importer before schwab_csv was added. If there's interest I can open source the Google equity awards source.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -28,7 +28,7 @@ appears in your Schwab CSV downloads. For example:
          schwab_account: "Brokerage XXXX-1234"
 
 You can also optionally specify accounts to be used for recording dividends, capital
-gains, fees, and taxes:
+gains, interest, fees, and taxes:
 
     2015-11-09 open Assets:Investments:Schwab:Brokerage-1234
          schwab_account: "Brokerage XXXX-1234"
@@ -60,7 +60,8 @@ download non-overlapping CSV statements.
 
 * Not all Schwab "actions" (transaction types) are supported. There's no reference for all
 the possible actions, and Schwab could add new ones any time. If your CSV includes an
-unsupported action, you'll get a `ValueError: 'Foo' is not a valid BrokerageAction`. Please
+unsupported action, you'll get a `ValueError: 'Foo' is not a valid BrokerageAction` or
+`ValueError: 'Foo' is not a valid BankingEntryType` for banking files. Please
 file an issue (and ideally a pull request!) to add support for that action.
 
 """
@@ -156,14 +157,11 @@ class BankingEntryType(enum.Enum):
 
 
 @dataclass(frozen=True)
-class RawBankEntry:
+class RawEntry:
     account: str
     date: datetime.date
-    entry_type: BankingEntryType
-    check_no: Optional[int]
     description: str
     amount: Optional[Amount]
-    running_balance: Amount
     filename: str
     line: int
 
@@ -172,13 +170,25 @@ class RawBankEntry:
 
     def get_processed_entry(
         self, account: str, account_meta: Meta
-    ) -> Optional[BankTransaction]:
+    ) -> Optional[TransactionEntry]:
+        raise NotImplementedError("subclasses must implement get_processed_entry")
+
+
+@dataclass(frozen=True)
+class RawBankEntry(RawEntry):
+    entry_type: BankingEntryType
+    check_no: Optional[int]
+    running_balance: Amount
+
+    def get_processed_entry(
+        self, account: str, account_meta: Meta
+    ) -> Optional[TransactionEntry]:
         interest_account = self.get_meta_account(account_meta,
                                             INTEREST_INCOME_ACCOUNT_KEY)
-        shared_attrs: SharedAttrsDict = dict(
+        shared_attrs: SharedAttrsDict = SharedAttrsDict(
             account=account,
             date=self.date,
-            entry_type=self.entry_type,
+            action=self.entry_type,
             description=self.description,
             amount=self.amount,
             filename=self.filename,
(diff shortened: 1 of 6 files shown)


Zburatorul:
Did you forget to include a commit with more code changes? I don't see the 7 new brokerage actions.

bayesianmind:
See line 134 and on. I didn't have to add new classes for them, they're similar enough to existing types.

beancount_import/source/schwab_csv.py
```diff
@@ -885,6 +1053,10 @@ def _get_key_from_posting(
             return None
         if not posting.account in account_set:
             return None
+        if not POSTING_META_ACTION_KEY in posting.meta:
+            return None
+        if not POSTING_DATE_KEY in posting.meta:
+            return None
```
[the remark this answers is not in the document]
carljm: If we do leave this in, I'd rather handle it the same as `source_desc` (get the value with an empty default and return `None` if it is empty, rather than checking if the key is there and then later getting the value for that key.)

beancount_import/source/schwab_csv.py
```diff
@@ -923,49 +1114,100 @@ def _load_transactions(filename: str) -> List[RawEntry]:
...
     return entries
 
 
+def _load_banking_transactions(reader: csv.DictReader, account: str, filename):
+    entries = []
+    transaction_start_line = 2
+    POSSIBLE_DATE=re.compile(r"[\d/\\]+")
+    for lno, row in enumerate(reader):
+        # First two rows are info messages.
+        if not POSSIBLE_DATE.match(row["Date"]):
```
carljm, line 1133: It feels a bit risky to me to silently skip any row with something in the date column that doesn't look to us like a date. Never know when Schwab might do something funky in that column (like in the brokerage side sometimes it looks like `DATE as of DATE`, which is handled by `_convert_date`), and when we see something unexpected, we should fail loudly and clearly, not silently skip transactions. Can we enumerate here the contents of the date column for the rows you want to skip, instead of skipping any row where it doesn't look like your regex?

(Edit: this intuition is confirmed by the fact that it looks like this is in fact currently causing us to silently skip a transaction in the misformatted test CSV file.)

testdata/source/schwab_csv/test_basic/transactions/EAC_Checking_Transactions_20201223-160009.CSV
```diff
@@ -0,0 +1,12 @@
+Transactions  for Checking account EAC as of 12/23/2020 16:00:10 ET
+"Date","Type","Check #","Description","Withdrawal (-)","Deposit (+)","RunningBalance"
+"Pending Transactions are not reflected within this sort criterion."
+"Posted Transactions""11/13/2020","TRANSFER","","Funds Transfer to Brokerage -9999","$10,500.00","","$50.01"
```
carljm, line 4: Something looks mis-formatted here. Is that coming directly from Schwab, or was accidentally introduced in this test file?
 
If the latter, let's re-verify the test file looks just like the CSV files you get from upstream.

If the former, that's unfortunate, but the importer needs adjusting to account for it so we don't skip the first transaction in every CSV. This row is currently silently ignored.

beancount_import/source/schwab_csv.py
```diff
@@ -923,49 +1114,100 @@ def _load_transactions(filename: str) -> List[RawEntry]:
...
+            entries = _load_banking_transactions(reader, account, filename)
+        else:
+            raise RuntimeError(f"Unexpected header {reader.fieldnames}")
     entries.reverse()
     return entries
 
 
+def _load_banking_transactions(reader: csv.DictReader, account: str, filename):
+    entries = []
+    transaction_start_line = 2
```
carljm, line 1129: It looks like all the source line numbers in the metadata for the generated transactions in the test are off by one (too high). Maybe this just needs to be `transaction_start_line = 1`?

beancount_import/source/schwab_csv.py
```diff
@@ -116,8 +117,9 @@
 )
 from beancount_import.unbook import group_postings_by_meta, unbook_postings
 
+CASH_CURRENCY="USD"
```
carljm, line 120: This is much better, thanks for pulling out this constant.

carljm (changes requested):
This looks great overall! Thanks so much for the additions and improvements. Made a few comments but overall I think this is pretty close to ready.

One thing is I noticed a few bad type annotations; haven't tested but I would think these should cause a mypy failure. Can you verify that `tox` is green with your changes?

beancount_import/source/schwab_csv.py
```diff
@@ -33,6 +33,7 @@
     2015-11-09 open Assets:Investments:Schwab:Brokerage-1234
          schwab_account: "Brokerage XXXX-1234"
          div_income_account: "Income:Dividend:Schwab"
+         interest_income_account: "Income:Interest:Schwab"
```
carljm, line 36: Oh, minor nit but can you also revise the prose sentence immediately preceding this example to read "... for recording dividends, interest, capital gains, fees, and taxes." (Addition of "interest," in the list.)

beancount_import/source/schwab_csv.py
```diff
@@ -59,7 +60,7 @@
 
 * Not all Schwab "actions" (transaction types) are supported. There's no reference for all
 the possible actions, and Schwab could add new ones any time. If your CSV includes an
-unsupported action, you'll get a `ValueError: 'Foo' is not a valid SchwabAction`. Please
+unsupported action, you'll get a `ValueError: 'Foo' is not a valid BrokerageAction`. Please
```
carljm, line 63: Could you also add a note here about the possibility of `ValueError: 'Foo' is not a valid BankingEntryType`? Unless you're confident that Schwab will never add any banking entry types you haven't already seen.

carljm:
I should also clarify that @jbms is the author and maintainer of beancount-import and the ultimate authority on whether this gets merged; just offering my review as the initial contributor of the Schwab CSV importer.

bayesianmind:
Thanks for the excellent review @carljm ! I will update this over the next few days as I have free cycles.

beancount_import/source/schwab_csv.py
```diff
@@ -1077,10 +1319,10 @@ def _load_positions_csv(
         symbol = STRIP_FROM_SYMBOL_RE.sub("", symbol)
         quantity = _convert_int(row["Quantity"])
```
zhou13, line 1320: The quantity of reinvested portfolios can be decimal. I need to replace this line with `quantity = _convert_decimal(row["Quantity"])`.

bayesianmind (commented):
Thank you both so much for your feedback. This PR is way better for it.
It's ready to go now. When combined with #96  tox is passing.

bayesianmind:
@jbms This is ready to merge alongside #96 when you get a chance. Thanks!

PR #98, opened 2021-02-17
ofx: support non-sale transfers out

FiloSottile:
This works with transactions like a charitable donation or a brokerage
transfer which don't involve liquidating the asset that is being transferred out.


jbms:
Thanks.  Can you by any chance add a test case for this?

Zburatorul:
@FiloSottile, I can add the test case if you can provide me with a sample block of OFX with the relevant case and the expected output.

Zburatorul:
It looks like this PR and #55 make overlapping and conflicting assumptions about `raw.unitprice <> None`.
Would appreciate some sample OFX @FiloSottile.

PR #101, opened 2021-02-28
Add promotional entries to schwab importer

zhou13:


jbms:
Thanks.  Can you by any chance add some tests?

carljm:
Should a promotional award go to the `interest_account`?

zhou13:
@jbms I will add some tests. @carljm In the US, promotional award generates 1099-INT for a brokerage account, which is interests for the purpose of tax.

carljm (changes requested):
@zhou13 In case it isn't clear, "adding tests" in this case should be as simple as adding a sample promotional entry row in the sample brokerage transactions file in `testdata/source/schwab_csv/test_basic/`. Then you can temporarily add `write=True` argument to the call to `check_source_example` in `test_source` in `beancount_import/source/schwab_csv_test.py`, run the tests once (`tox` should do it), verify that the modified test output files look correct, remove the `write=True` arg, and then commit the resulting changes to the tests. Let me know if you have any trouble with this process.

Zburatorul:
Carl, the test writing guidance you shared is useful. I've been simply running `pytest -k schwab`, looking at the errors for the diff, and then manually applying that diff to the right files. Your way is less work.

zhou13:
Thanks for the guide. I have added the test. I also include another entry `PRIOR_YEAR_SPECIAL_DIVIDEND` in the commit.

testdata/source/schwab_csv/test_basic/import_results.beancount
```diff
@@ -582,6 +582,28 @@
 
 2020-11-15 price SMAL040120203200P                   12.53 USD
 
+;; date: 2020-11-19
+;; info: {"filename": "<testdata>/test_basic/transactions/Brokerage_Transactions_20201115-180021.CSV", "line": 3, "type": "text/csv"}
+
+; features: []
+2020-11-19 * "INTEREST - PROMOTIONAL AWARD"
```
Zburatorul, line 589: Nit comment, but can you add an if in the BankInterest class so the narration prefix for promo award is not INTEREST, but something else. Probably INCOME?

zhou13: Done.

carljm (approved):
This looks good to me!

carljm:
@zhou13 Sorry for the hassle, but unfortunately another PR that modified the same files in conflicting ways was merged before this, and this now requires updates to be mergeable. Can you rebase on master (or merge in master), resolve the conflicts, and push an updated version of this PR?

carljm:
Thanks!

PR #103, opened 2021-03-29
[schwab] parse lot-details CSV to fill in cost on stock buy/sell

carljm:
The current version of the Schwab CSV importer in beancount-import has a couple TODO items related to stock sales and purchases. On a stock purchase we currently have a "1 FIXME" for the lot cost, and on stock sales we currently aren't able to fill in cost-basis information for any of the stock sold. Both of these then require manual fixup after import.

This PR resolves both TODOs by (optionally) parsing the lot-details CSV from Schwab (which the finance-dl downloader is capable of downloading at every update) in order to figure out the relevant cost-basis information for both situations. Assuming we have a lot-details update between each transation, we can reliably get the correct cost-basis lots for each sale, including a single sale that sells stock from multiple lots.


Zburatorul:
Carl, is it necessary to download the lot details from Schwab? Why can they not be inferred from the Beancount journal file?

carljm:
The Schwab transactions CSV doesn't report how many units from which existing lots were sold in a given sale. There's no way this information can be inferred. Without passing the Schwab lot details csv to figure it out, I have to do the equivalent manually when importing each sale.

bayesianmind:
This is clever. Reviewed this PR and the approach it takes it looks solid to me.

bayesianmind (approved):

Zburatorul:
> The Schwab transactions CSV doesn't report how many units from which existing lots were sold in a given sale. There's no way this information can be inferred without more information from Schwab.

The cost basis of every lot is present in the journal file, assuming those transactions have already been imported.
There are only a handful of ways the lot in a sale can be chosen: fifo, lifo, average. Beancount accounts can have a specific booking method declared. So one could get a list of all lots of a given commodity and match according to the rule. I don't know how to do this, but I'm guessing some beancount method must get close.

>This PR resolves both TODOs by (optionally) parsing the lot-details CSV from Schwab (which the finance-dl downloader is capable of downloading at every update) in order to figure out the relevant cost-basis information for both situations. Assuming we have a lot-details update between each transation, we can reliably get the correct cost-basis lots for each sale, including a single sale that sells stock from multiple lots.

The lot downloader fails for me while matching the `table.securityTable tr.data-row` elements. Schwab might have changed the HTML.
Looking at the Positions CSV I see a Cost Basis column in it. Seems the cost basis can be derived without the need to download anything.

carljm:
@Zburatorul 

> There are only a handful of ways the lot in a sale can be chosen: fifo, lifo, average. Beancount accounts can have a specific booking method declared. So one could get a list of all lots of a given commodity and match according to the rule.

Unfortunately I don't think it is necessarily that simple. For example, Schwab offers at least seven different options (http://www.schwabadvisorcenter.com/cms/P-3831954.1/Cost_Basis_Methods_Final_12_2_10.pdf?cmsid=P-3831954&cv16 ), including "tax lot optimizer" which has a more complex algorithm, not to mention arbitrary choice of lot via contacting an adviser. I don't think it is realistic to try to locally mirror every lot choice algorithm that might be available; it is more robust to use the information that Schwab makes available to figure out how the transaction was actually handled, if possible.

That said, there's a reason I've made this an optional feature! If it isn't necessary or useful for your use case, you can just not use it. I'm also not opposed to reviewing pull requests for additional or alternative cost-basis heuristics for simpler situations.

> Looking at the Positions CSV I see a Cost Basis column in it. Seems the cost basis can be derived without the need to download anything.

Yes, I also saw this, but this is an aggregate cost basis for all lots, and I don't think in general it is possible to know simply from the change in this aggregate exactly how many units from each lot were sold. Or maybe it is possible, if we have correct information for all existing lots, and all lots have distinct unit costs? It seems plausible that it is possible, but I'm not entirely sure and I haven't worked through the math :) If I see a PR that successfully handles it that way and renders this approach redundant, I'll be happy to remove this in favor of that!

@bayesianmind 

> This is clever. Reviewed this PR and the approach it takes it looks solid to me.

Thanks for reviewing! I'm not ignoring your PR that you requested my review on, just haven't had a moment yet; I should be able to review it tonight or this weekend.

carljm:
Oh, and

> The lot downloader fails for me

I ran it just now successfully. One recent issue that could be affecting you is that Schwab added an irritating "are you looking for your tax documents?" modal dialog. I was able to get past that by manually clicking "Don't show this again." I assume after tax season that dialog will go away anyway.

Zburatorul:
@carljm, thanks for the explanation regarding cost basis. I see now it is indeed trickier than I thought.
> Oh, and
> 
> > The lot downloader fails for me
> 
> I ran it just now successfully. One recent issue that could be affecting you is that Schwab added an irritating "are you looking for your tax documents?" modal dialog. I was able to get past that by manually clicking "Don't show this again." I assume after tax season that dialog will go away anyway.

No, I don't see any dialogs. It loads the "Market Value" tab and then just sits there until failure on line 209.
Can you tell me how to access this information myself in the UI please? I should be able to diagnose it once I know what I'm supposed to see.

carljm:
@Zburatorul Sorry I didn't get back to this sooner, but it seems you found the lot details in the UI in the meantime.

carljm:
Hi @jbms, do you have any thoughts on this PR, or do you think it is OK to merge? It has an approving review from @bayesianmind.

PR #104, opened 2021-04-06
Schwab: Added WIRE and SECURITY_TRANSFER support

bayesianmind:
Security transfers almost always require rewriting to capture the original
lots. Due to that, we need a way to match the original transaction. The
best way would be to match the original CSV line entirely, but that's
not the design of the importer. Instead, we now have a "schwab_amount"
metadata override you can (on one posting) to dedupe the original CSV
entry on.

Passes tox.


bayesianmind:
@carljm Would love if you could review this

beancount_import/source/schwab_csv.py
```diff
@@ -145,15 +145,19 @@ class BrokerageAction(enum.Enum):
...
     WIRE_FUNDS_RECEIVED = "Wire Funds Received"
     MISC_CASH_ENTRY = "Misc Cash Entry"
+    JOURNALED_SHARES = "Journaled Shares"
+    SECURITY_TRANSFER = "Security Transfer"
+    EXERCISE = "Exchange or Exercise"
 
 
 class BankingEntryType(enum.Enum):
     ACH = "ACH"
+    ATM = "ATM"
```
bayesianmind, line 155: I just sorted this list

carljm:
Ok, took a look! Not sure what's up with the test failures, need to get that cleared up. The new transaction types (and testing of them) looks good.

I'm not sure about the new `amount` key. It seems to be solving a problem that I don't understand or don't have, and I'd like to understand what is different about our setups that is causing you this problem. I regularly split stock sale transactions into multiple postings for different cost basis lots (previously by hand, and more recently via automation in my PR for parsing Schwab lot details CSV). My experience has been that as long as all the postings have the right `date`, `schwab_action`, and `source_desc` metadata, beancount-import has never had any problem recognizing them as cleared.

One option here is to reduce this PR to just the new transaction types so we can get that merged right away while we debug the other situation.

bayesianmind:
re build failures: #106 fixes some missing node deps that were making tox fail for me locally.

re amount: The core problem is that `_get_key_from_posting` includes the exact Amount of the posting as a key (via `final_units`).
I split the generated posting up to match lots, either for a sale covering multiple lots, or for transferring lots to a new broker.

For example consider the following Sale:
```
2021-02-23 * "SELLSTOCK - ALPHABET INC. CLASS C"
  Assets:Schwab:Brokerage:GOOG                       -0.322 GOOG {1343.56 USD} @ 2052.0553 USD
  Assets:Schwab:Brokerage:GOOG                       -0.678 GOOG {1295.34 USD} @ 2052.0553 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
    schwab_amount: "-1 GOOG"
  Assets:Schwab:Brokerage:Cash                     2052.04 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
  Income:Schwab:Brokerage:Gains:GOOG
  Expenses:Brokerage-Fees:Schwab                       0.01 USD
```

The original generated posting would have just been:
```
  Assets:Schwab:Brokerage:GOOG                       -1 GOOG {} @ 2052.0553 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
```

Since I needed to tell Beancount which lots Schwab selected (I use the tax optimizer for lot selection), I have to split the posting. In Beancount V3 I hope to work on configurable booking that would allow user implemented booking so we make Beancount automatically choose the same lots that Schwab does.

.gitignore
```diff
@@ -6,6 +6,7 @@ frontend/node_modules
 beancount_import/frontend_dist/*.js
 beancount_import/frontend_dist/*.css
 beancount_import/frontend_dist/*.map
+beancount_import/frontend_dist/dev
```
jbms, line 9: I think this addition should not be necessary anymore?

bayesianmind: Yes good call, removed it and added `.eggs/`

carljm:
@bayesianmind So the problem you are describing with splitting postings for booking should already be solved by the use of `group_postings_by_meta` and then `unbook_postings`. What this does is group postings in a transaction that have matching meta values, and then combine them back into a single posting for purposes of matching cleared postings.

I also split postings for booking in exactly the way you show, and the existing code to re-group these and match them works for me, but apparently not for you. So I'd like to understand why before we add a new separate workaround.

Is it perhaps because you are splitting fractional shares, and due to floating point inaccuracies you end up with a re-grouped total that is very close but doesn't quite match the original amount? Perhaps this could be solved by rounding the amount to within a certain tolerance for generating the key?

carljm:
@bayesianmind On a second closer look, it is clear why you are having this problem. You aren't duplicating the posting metadata on both splits. By not doing that, you've lost the connection to the original transaction and the grouping/unbooking code doesn't realize that these should be re-grouped for matching to the original transaction.

In other words, your split transaction should look like this:

```
2021-02-23 * "SELLSTOCK - ALPHABET INC. CLASS C"
  Assets:Schwab:Brokerage:GOOG                       -0.322 GOOG {1343.56 USD} @ 2052.0553 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
  Assets:Schwab:Brokerage:GOOG                       -0.678 GOOG {1295.34 USD} @ 2052.0553 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
  Assets:Schwab:Brokerage:Cash                     2052.04 USD
    date: 2021-02-23
    schwab_action: "Sell"
    source_desc: "ALPHABET INC. CLASS C"
  Income:Schwab:Brokerage:Gains:GOOG
  Expenses:Brokerage-Fees:Schwab                       0.01 USD
```

If you do that, the grouping/unbooking code will work as designed (re-combining those two postings because their account and metadata matches) and it will match with the original transaction, with no need for the new `schwab_amount` key.

carljm (changes requested):
Per comment above, I think the new `schwab_amount` stuff should be removed from this PR, and then we can merge just the two new transaction types.

bayesianmind:
Ahhh thanks @carljm! This is much better and less hacky. Good to learn about `group_postings_by_meta`.
I pushed those changes.

carljm (approved):
Looks good!

bayesianmind:
@carljm Turns out there are bugs in `group_postings_by_meta` that I had to fix before it would work for me as you've described. I would've expected you to hit the same bug. I run against beancount v3, this may be a difference in v3?
My fix is in #110.

bayesianmind:
Ah jbms gave me more context - if the postings are split automatically by beancount itself (for example by the FIFO booker), then entries will have the same lineno.

carljm (approved):
LGTM!

PR #105, opened 2021-04-09
Multiple new Schwab transaction types

Zburatorul:
This adds support for five new transaction types.

The way I dealt with `STOCKMERGER` by adding new fields to `BankingEntryType` and modifying the loop in `_load_brokerage_transaction` is ugly. Better suggestions welcome.
Furthermore, it is unclear what cost basis to use for the lot that's being reduced so I insert "FIXME" instead which at least gets the transaction in the journal and prompts the user to choose a lot manually.


beancount_import/source/schwab_csv.py
```diff
@@ -1113,10 +1295,17 @@ def _load_brokerage_transactions(reader: csv.DictReader, account: str,
...
+                merger_quantity=merger_quantity,
+                merger_description=merger_desc,
                 filename=filename,
                 line=lno + 2,
             )
         )
+        if merger_symbol is not None:
+            merger_symbol = None
+            merger_quantity = None
+            merger_desc = None
```
bayesianmind, line 1308: If you wanted to avoid this one idea is to modify the loop like I do here: https://github.com/bayesianmind/beancount-import/blob/master/beancount_import/source/schwab_equityaward.py#L108

I think the way you've done it here is the least bad option FWIW.

beancount_import/source/schwab_csv.py
```diff
@@ -219,6 +224,9 @@ class RawBrokerageEntry(RawEntry):
     quantity: Optional[Decimal]
     price: Optional[Decimal]
     fees: Optional[Decimal]
+    merger_symbol: Optional[str]
+    merger_quantity: Optional[Decimal]
+    merger_description: Optional[str]
```
bayesianmind, line 229: This isn't ideal as you note.
It's beyond the scope of this PR, but given Schwab's extremely limited CSV format requiring them to do ugly things like break a single transaction into multiple rows it is better to have flexibility when parsing. If we move away from the thin simple parser that just dumps CSV rows into `RawBrokerageEntry` and instead move the logic of `get_processed_entry` into the CSV reading logic it would be a net improvement (in this programmer's opinion). If we had that structure you'd construct the Merger data class in the CSV parsing loop.

The approach you take here isn't terrible though and I would be fine with it if it were my codebase. If we ended up finding several more transaction types like this then it will get really ugly, but I think it's unlikely there are too many transaction types that span multiple entries like this.

bayesianmind (approved):

beancount_import/source/schwab_csv.py
Zburatorul: I think we should defer the act moving the logic you mention into the CSV reading logic until another such situation arises. It will at least be more clear how to do it then.

bayesianmind: Makes sense to me.

carljm: I'm not sure it really helps to move more logic into CSV parsing. Either way we end up having to do the same thing to handle a case like this (loop over sequential rows with some aggregation logic). The intention of the design is that the `Raw*` objects should be a very close mirror of exactly what is in the CSV file, and then in a separate step we process those raw entries into something more useful for importing into beancount. This separates the CSV-parsing details from the logical interpretation of the CSV rows actually mean. To me this seems like useful separation of concerns, and I don't see what is gained by collapsing them into a single step.

beancount_import/source/schwab_csv.py
```diff
@@ -307,6 +321,38 @@ def get_processed_entry(
...
+                action=self.action,
+                description=self.description,
+                quantity=self.quantity,
+                symbol=self.symbol,
+                price=price,
+                fees=self.fees,
+                amount=amount,
+                filename=self.filename,
+                line=self.line,
+            )
```
carljm, line 342: rather than duplicating the creation of the shared_attrs dict, let's follow the existing pattern of `STOCK_PLAN_ACTIVITY` which also needs to do some prior munging of `amount`, and have two `if self.action is BrokerageAction.EXPIRED`, one before the construction of `shared_attrs` and one after.

beancount_import/source/schwab_csv.py
```diff
@@ -307,6 +321,38 @@ def get_processed_entry(
...
+
+        if self.action in (BrokerageAction.SHORT_TERM_CAP_GAIN, BrokerageAction.LONG_TERM_CAP_GAIN):
+            return FundGainsDistribution(symbol=self.symbol, capital_gains_account=capital_gains_account, **shared_attrs)
+
+        if self.action == BrokerageAction.STOCK_MERGER:
+            assert self.quantity is not None
+            return Merger(account=account, fees_account=fees_account, symbol=self.symbol, quantity=self.quantity, price=self.price,
+                          amount=self.amount, merger_symbol=self.merger_symbol, merger_quantity=self.merger_quantity,
+                          date=self.date, action=self.action, description=self.description, fees=self.fees, filename=self.filename,
+                          merger_description=self.merger_description, line=self.line)
```
carljm, line 354: again here it looks like we are re-handling things that are already handled in `shared_attrs`, and then not using `shared_attrs`. This makes future maintenance more difficulty; if we need to make global adjustments that apply to a certain shared attribute of all transaction types, it will appear that we can make this change in `shared_attrs`, and we may not notice that some transaction types aren't using it. So I'd rather use the existing pattern of `STOCK_PLAN_ACTIVITY` as mentioned above to ensure that all transaction types actually do share `shared_attrs`. If we have to use that pattern too many times for some field, we can move that field out of `shared_attrs` and instead pass it individually for each transaction type.

beancount_import/source/schwab_csv.py
```diff
@@ -531,6 +591,44 @@ def get_narration_prefix(self) -> str:
...
+                flag=None,
+                meta=self.get_meta(),
+            ),
+            Posting(
+                account=self.get_other_account(),
+                units=None,
+                cost=None,
+                price=None,
+                flag=None,
+                meta=self.get_meta(),
```
carljm, line 628: We could introduce a `get_other_meta()` method (defaulting to `return {}` but overridden here to `return self.get_meta()`) so that you don't have to override all of `get_postings()` and repeat a lot of what is already done in the base method. It looks like all the other differences here can already be handled by existing methods e.g. `get_other_units()` and `get_cost()`.

I don't feel strongly about this though, this override is OK.

beancount_import/source/schwab_csv.py
carljm: I agree, this doesn't seem bad at all to me. I think it makes sense in this case to do this line merging here in the CSV parsing, even though it's a slight departure from the "RawEntries are 1:1 representations of rows in the CSV" design.

One small tweak I might make is to store the three items as a tuple or NamedTuple, so that you can just have a single variable to set to either None or that tuple, and then not have to arbitrarily choose one of them to check for None-ness.

carljm (commented):
Looks good, thanks! A few small suggestions, but the handling of mergers seems fine overall.

beancount_import/source/schwab_csv.py
Zburatorul: Introduced the nametuple.

beancount_import/source/schwab_csv.py
Zburatorul: Done.

beancount_import/source/schwab_csv.py
Zburatorul: Done.

beancount_import/source/schwab_csv.py
Zburatorul: Eh, I prefer less code so I removed the override.

Zburatorul:
I have made the changes suggested by the latest comments and this is ready to merge, unless more objections.

carljm (approved):
Looks good to me! Fine with me to merge it now and update with lot info later, or wait to merge and update it, either way.

Zburatorul:
Rebased and got tox to pass. Ready to merge.
Will get it to work with latest lot capabilities in separate PR.

carljm (changes requested):
A few review comments, mostly nits that I'm fine with either way, but the duplicate `EXPIRED` blocks and the changes in `test_lots/` should be fixed.

testdata/source/schwab_csv/test_lots/positions/Intelligent-Positions-2020-10-25.CSV
```diff
@@ -1,6 +1,6 @@
 "Positions for account Intelligent XXXX-4321 as of 02:45 AM ET, 10/25/2020"
 
 "Symbol","Description","Quantity","Price","Price Change $","Price Change %","Market Value","Day Change $","Day Change %","Cost Basis","Gain/Loss $","Gain/Loss %","Reinvest Dividends?","Capital Gains?","% Of Account","Dividend Yield","Last Dividend","Ex-Dividend Date","P/E Ratio","52 Week Low","52 Week High","Volume","Intrinsic Value","In The Money","Security Type",
-"SCHA","SCHWAB US SMALL CAP ETF","10.5","$70.50","$0.00","0.00%","$740.25","$0.00","0.00%","$759.68","-$8.50","-0.1%","No","--","5.24%","+1%","$0.30","8/10/2020","--","$43.05","$89.52","431,291","--","--","ETFs & Closed End Funds",
+"SCHA","SCHWAB US SMALL CAP ETF","10","$70.50","$0.00","0.00%","$705.00","$0.00","0.00%","$713.50","-$8.50","-0.1%","No","--","5.24%","+1%","$0.30","8/10/2020","--","$43.05","$89.52","431,291","--","--","ETFs & Closed End Funds",
```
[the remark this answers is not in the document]
Zburatorul: Indeed. Fixed.

beancount_import/source/schwab_csv.py
```diff
@@ -587,6 +679,24 @@ def get_narration_prefix(self) -> str:
         return "INTEREST"
 
 
+@dataclass(frozen=True)
+class FundGainsDistribution(TransactionEntry):
+    '''
+        ETFs and Mutual Funds can have distributions of capital gains
+        generated by internal activity.
+    '''
```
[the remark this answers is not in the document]
Zburatorul: Done.

beancount_import/source/schwab_csv.py
```diff
@@ -356,6 +410,30 @@ def get_processed_entry(
             return TaxPaid(taxes_account=taxes_account, **shared_attrs)
         if self.action == BrokerageAction.MARGIN_INTEREST:
             return MarginInterest(**shared_attrs)
+        if self.action is BrokerageAction.EXPIRED:
```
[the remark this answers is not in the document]
Zburatorul: Rebase conflict artifact.

beancount_import/source/schwab_csv.py
```diff
@@ -279,6 +290,11 @@ def get_processed_entry(
             filename=self.filename,
             line=self.line,
         )
+        if self.action == BrokerageAction.STOCK_MERGER:
+            assert self.quantity is not None
+            assert self.merger_spec is not None
+            return Merger(fees_account=fees_account, symbol=self.symbol, quantity=self.quantity, price=self.price, fees=self.fees,
+                          merger_spec=self.merger_spec, **shared_attrs)
```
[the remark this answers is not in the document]
Zburatorul: Changed for consistency.

beancount_import/source/schwab_csv.py
```diff
@@ -176,8 +176,15 @@ class BrokerageAction(enum.Enum):
     MISC_CASH_ENTRY = "Misc Cash Entry"
     JOURNALED_SHARES = "Journaled Shares"
     SECURITY_TRANSFER = "Security Transfer"
+    EXPIRED = "Expired"
+    STOCK_MERGER = "Stock Merger"
+    BOND_INTEREST = "Bond Interest"
+    SHORT_TERM_CAP_GAIN = "Short Term Cap Gain"
+    LONG_TERM_CAP_GAIN = "Long Term Cap Gain"
```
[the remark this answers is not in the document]
Zburatorul: Let's do that.

beancount_import/source/schwab_csv.py
```diff
@@ -176,8 +176,15 @@ class BrokerageAction(enum.Enum):
...
 
 
+MergerSpecification = namedtuple('MergerSpecification', ['symbol', 'quantity', 'description'])
```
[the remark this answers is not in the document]
Zburatorul: Changed it to a dataclass. One odd thing is that quantity must be Optional[Decimal] for mypy to pass (because that's what _convert_decimal returns), but in reality it really cannot be None. This means in such situations we end up inserting asserts right before the creation of the dataclass. I don't know of a better pattern. Let me know if you do.

beancount_import/source/schwab_csv.py
```diff
@@ -346,6 +378,31 @@ def get_processed_entry(
                 fees=self.fees,
                 **shared_attrs,
             )
+        if self.action is BrokerageAction.EXPIRED:
+            assert self.quantity is not None
+            amount = Amount(Decimal(0), currency=CASH_CURRENCY) if self.amount is None else self.amount
+            price = Decimal(0) if self.price is None else self.price
+            lot_info = lots.get_sale_lots(account_meta["schwab_account"], self.symbol, self.date, self.quantity)
```
carljm, line 385: This `lot_info` variable is not used.

beancount_import/source/schwab_csv.py
```diff
@@ -346,6 +378,31 @@ def get_processed_entry(
                 fees=self.fees,
                 **shared_attrs,
             )
+        if self.action is BrokerageAction.EXPIRED:
```
carljm, line 381: very minor nit, but the rest of the cases in this method use `==` not `is`. I think both should work equally well in this case but we may as well be consistent

beancount_import/source/schwab_csv.py
```diff
@@ -346,6 +378,31 @@ def get_processed_entry(
...
+            assert self.quantity is not None
+            amount = Amount(Decimal(0), currency=CASH_CURRENCY) if self.amount is None else self.amount
```
carljm, line 383: I think the wrong version of this block was kept after merge conflict, and we've now lost some of my earlier feedback. Notably we are redefining `amount` here after already having defined it above, and we are not using `**shared_attrs` below.

beancount_import/source/schwab_csv.py
```diff
@@ -766,6 +859,76 @@ def get_accounts(self) -> List[str]:
...
+        postings = [
+            Posting(
+                account=self.get_primary_account(),
+                units=Amount(self.quantity, currency=self.symbol),
+                cost=CostSpec(
+                    number_per=self.price,
+                    number_total=None,
+                    currency=CASH_CURRENCY,
+                    date=None,
+                    # at the moment requires manually choosing the lot
```
carljm, line 887: I think this could be resolvable via `lot_info`, but it's OK to merge without implementing that yet.

carljm (changes requested):
Sorry, saw a few things yet in this updated version.

beancount_import/source/schwab_csv.py
Zburatorul: Good catch.

beancount_import/source/schwab_csv.py
Zburatorul: I want to handle this in a separate PR. This PR is big as is.

beancount_import/source/schwab_csv.py
Zburatorul: Changed

beancount_import/source/schwab_csv.py
Zburatorul: Passed it to Sell().

Zburatorul: Anything to do with lots I'm leaving for future PRs.

testdata/source/schwab_csv/test_lots/positions/Intelligent-Positions-2020-10-25.CSV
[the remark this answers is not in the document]
carljm: It seems like these changes in test_lots are still here? I made this 10.5 intentionally to validate that the lot details handling allows for fractional shares.

carljm (approved):
Looks good, thank you!!

PR #106, opened 2021-04-10
Add missing node dependencies

bayesianmind:
Tox fails locally for me on 2 different machines without these deps.


bayesianmind:
Hm I can't repro this anymore. Maybe it's not needed after all.

PR #52, opened 2020-04-03
Add Missing Cash Balance Entries for Investments

patbakdev:
Add a cash balance directive for investment accounts based on the
available cash element using the first matching dated balance entry.

Added additional tests for fidelity.ofx

PR #107, opened 2021-04-14
Add Missing Cash Balance Entries for Investments

patbakdev:
This is a second attempt at Add Missing Cash Balance Entries for Investments #52

Add a cash balance directive for investment accounts based on the
available cash element using the first matching dated balance entry.

Added additional tests for fidelity.ofx


patbakdev:
Tests pass locally (Linux; Python 3.9). Not sure what these failures (failed to build frontend bundles) are due to. Seems to be a common problem. Let me know what else I need to do.

jbms:
Thanks.

PR #108, opened 2021-04-14
Add Missing Security ID to Error List Instead of Throwing

patbakdev:
This was a quick fix I did some time back. Still seems relevant.


patbakdev:
Tests pass locally (Linux; Python 3.9). Not sure what these failures (failed to build frontend bundles) are due to. Seems to be a common problem. Let me know what else I need to do.

PR #110, opened 2021-04-19
Allow group_postings_by_meta to work for manually booked lots.

bayesianmind:
group_postings_by_meta is used in description_based_source and schwab_csv. Both iterate over journal entries which always have file and lineno meta set. Since each posting has a unique lineno, this method won't group postings that it should.

Additionally, postings with different accounts shouldn't be grouped together. I've added that condition as well. That bug was probably masked previously by the first bug.

I've copied the post metadata, otherwise we'd be modifying the input object and beancount-import depends on lineno meta downstream.


jbms:
To be clear, the issue this function was originally intended to solve is that if your journal contains a transaction like:

2021-01-01 * Whatever
  Assets:Bank:TICKER   -20 TICKER {} @ 5 USD
  Income:Bank:Capital-Gains:TICKER
  Assets:Bank:Cash 100 USD
  
Then if the 20 TICKER sold come from more than one lot, the booking procedure will expand this into:

2021-01-01 * Whatever
  Assets:Bank:TICKER   -15 TICKER {4 USD, 2020-05-05} @ 5 USD
  Assets:Bank:TICKER   -5 TICKER {3 USD, 2019-04-04} @ 5 USD
  Income:Bank:Capital-Gains:TICKER -30 USD
  Assets:Bank:Cash 100 USD
 
Having two postings in place of one causes problems with checking for duplicates / invalid entries, so unbook.py converts them back into a single entry.

But those two postings should indeed still have the same line number, since they are derived from the same source line.

It sounds like you want to use this function for a different purpose?

bayesianmind:
Ah I see. I'm trying to solve the same problem but when the postings are split manually rather than FIFO.
e.g the user split the -20 TICKER themselves by breaking it into 2 lots.
FIFO booking method doesn't always choose the same lots as the broker so this is very nice to have.

Since the intent is the same, and it won't break the existing usecase it seems reasonable to me to expand it slightly.

jbms:
I think if you are splitting manually, it will already work if you just keep the import metadata on only one of the postings, and don't copy it to the other postings.

carljm:
> I think if you are splitting manually, it will already work if you just keep the import metadata on only one of the postings, and don't copy it to the other postings.

Heh, I just told @bayesianmind the opposite! It's always worked fine for me _with_ copied metadata, and doesn't work without copying the metadata. But it could be true that I'm on an older version of beancount still. I'll check on this later and see if I can figure out why it's working for me with manually split postings with copied metadata.

But I don't think it was working for @bayesianmind without copying the metadata to all postings, either...

bayesianmind:
The `schwab_csv` source uses posting Amount as part of its dudupe key mechanism - so if a posting is split (either by FIFO or manually for SpecID booking) then the dedupe fails to match. That's why the schwab_csv source uses `group_postings_by_meta` and `unbook_postings`.

carljm:
Ok I hadn't looked closely enough at how `group_postings_by_meta` works and was spreading wrong info, sorry @bayesianmind! Since it checks identity of the meta dictionary, it definitely currently only works for unbooking auto-booked postings, not for postings that are explicitly booked in the journal. With the current code, the latter cannot work, whether the meta is duplicated or not. If you duplicate the meta, the meta dictionary will still have a different identity (not to mention different line number, though that's irrelevant since dictionary identity, not equality, is checked), so the postings won't be grouped. If you don't duplicate the meta, the extra postings will have meta `None`, and thus still won't be grouped.

It works for me now because I use the lot-details parsing in the Schwab CSV importer, so the importer splits the posting for booking according to how it was actually booked by Schwab (not using beancount auto-booking), and it will split it the same way on all future imports too, so all the postings in the journal always match the ones from the importer. If I remove the lot-details CSV files from my importer config for the Schwab source, my previously split booked postings do indeed cause errors due to partially-matching transaction.

I used to manually split postings for booking, and I don't remember ever having problems with clearing. But I think maybe I only ever did that back when I was still using OFX import, and the OFX importer behaves differently since it can reliably match by `ofx_fitid`, so it doesn't even need to use `unbook_postings()`, and I never had this problem.

As far as how to address this going forward: to me it makes sense that the Schwab importer at least should be able to match up explicitly booked trades, even if you aren't using the lot details parsing and are doing the booking manually. (Though I recommend trying the lot-details parsing now that it's merged!) I can see a few ways for that to happen:

1) Change the behavior of `group_postings_by_meta` as this PR does, so that it groups by meta equality without considering line number, instead of grouping by meta identity. I think this is reasonable and safe behavior; it will still work for auto-booked postings as it does today, and will also work for explicitly booked postings.

2) We could add a new meta key (something like `merge: true`, though I don't really care what it's called) to explicitly indicate "merge with previous posting", and `group_postings_by_meta` could just gain a single new special case where if `merge: true` is found in the meta of a posting, it merges that posting as if it had identical meta to the previous posting. Then instead of duplicating meta when manually splitting postings for booking, you'd just leave the meta on the first posting and add only `merge: true` to all the rest. I think this is also a fine option: it really just depends whether we prefer duplicated meta or single new meta key on the extra postings. I personally kind of like duplicated meta since it clearly communicates that the postings all came from the same original posting (and because that's how the lot-details auto-splitting already does it), but I could be fine with this option too.

If @jbms would rather keep the behavior of `group_postings_by_meta` exactly as-is by default, we could also add a new flag that can be passed to it to enable one of the above behaviors opt-in for just the Schwab importer. (Or just have our own copy of `group_postings_by_meta` in the Schwab importer.)

Sorry for the novel, hope that was all clear. Thoughts?

bayesianmind:
Thanks @carljm! Looking forward to trying out the lot details parser!
I agree with those options. Option 1 seems quite a lot cleaner. This PR implements that, I'll add a unit test additionally.

beancount_import/unbook.py
```diff
@@ -5,20 +5,32 @@
...
     posting_list = []
+    postings = sorted(postings, key=lambda posting: posting.account)
     for posting in postings:
-        if posting.meta is not None and posting.meta is prev_meta:
+        # We need to ignore lineno for comparison purposes, otherwise this
+        # method will never book any postings.
+        post_meta = posting.meta.copy()
+        # Parser meta `filename` and `lineno` metadata should be ignored.
+        if "filename" in post_meta:
+            del post_meta["filename"]
```
carljm, line 23: Do we really want to delete filename? I would think postings split for booking must still always be in the same transaction, therefore same file. Shouldn't matter though, since this function should never be called with postings not from the same transaction.

carljm (commented):
@bayesianmind So I just realized that this approach may not work, and we may need to go for option 2 instead (the explicit `merge me` metadata for manually split postings that should be merged). The reason is that I think the current form of this would actually break postings split for booking by the importer itself (via lot details); we'd end up merging the postings on the transaction from the journal, but the transaction from the importer would still have split postings, and they won't match. In general this will break matching any time an importer actually intentionally creates transactions that have multiple postings with same account and same metadata; which the Schwab importer (with lot details) does.

If we do go with option 2 instead, curious for @jbms ' thoughts on whether this seems generally useful, or should be a feature specific to the Schwab importer.

beancount_import/unbook.py
bayesianmind: I added it so that postings with no meta on them won't get merged.
Since we are using meta equality instead of identity two postings with no meta on them would otherwise get grouped.

bayesianmind:
> @bayesianmind So I just realized that this approach may not work, and we may need to go for option 2 instead (the explicit `merge me` metadata for manually split postings that should be merged). The reason is that I think the current form of this would actually break postings split for booking by the importer itself (via lot details); we'd end up merging the postings on the transaction from the journal, but the transaction from the importer would still have split postings, and they won't match. In general this will break matching any time an importer actually intentionally creates transactions that have multiple postings with same account and same metadata; which the Schwab importer (with lot details) does.
> 
> If we do go with option 2 instead, curious for @jbms ' thoughts on whether this seems generally useful, or should be a feature specific to the Schwab importer.

Good point. A simple fix could be running unbooking on both the import result and the journal (just for the purposes of duplicate suppression). Then if the importer had generated multiple postings they'd be merged in the same way as the journal and be correctly suppressed as dupes.
I'll make sure to test this with your lot selector.

I really don't prefer option 2, it doesn't seem very extensible or elegant.

carljm:
@bayesianmind The nice thing about option 2 is that it is very explicit. I agree it doesn't feel as elegant.

I'm open to your fix, but if we go that route it leads me to think the changes should perhaps be isolated to the Schwab importer (in other words, have our own function for merging postings for unbooking, and leave this function for use where the only postings that should be merged are those that were split by beancount's automated booking procedure.) I'm just a little concerned about what cases there could be in other importers (current or future) that we don't know about that could be similarly broken by this behavior.

In the Schwab importer specifically, I'm comfortable with your fix because I know what metadata appears on all transactions from the importer, and I'm confident that postings with duplicate account and metadata really can only occur as a result of explicit lot selection, so the merging behavior is correct and fine. I don't have the same confidence in general for all potential users of `group_postings_by_meta`.

jbms:
I think it would be helpful to keep the different sources consistent in how manual splits should be handled, i.e. whether you are supposed to duplicate the metadata or not.

For the OFX source, you are not supposed to duplicate the metadata.  So, if it is possible to just make that work with the schwab source, that would seem like the simplest solution, since then we retain consistency.

However, I can see that it poses some additional difficulties given the lack of unique transaction ids.

As you note, just duplicating the metadata without any special `merge` indication is problematic since then it is ambiguous whether those postings were split or actually were derived from separate source data that happens to have the same description, etc.

Could the schwab source be changed to not use the amount at all for matching against imported data?  Already it is not guaranteed to be unique anyway, but I suppose might prevent certain issues from being detected that would currently be detected by a mismatch in the amount.

Then it would also be necessary for schwab users to fix their journals to unduplicate the metadata, but perhaps that would not be too many transactions.

jbms:
I also agree with you that having some explicit indication that the postings are related, like `merged: true`, does have some advantages.

bayesianmind:
Good points. I wish that Schwab didn't remove their OFX endpoint, or that they'd let us build against their real APIs. I applied as a dev in Nov 2020 after they killed their OFX gateway.

Regarding `merge: true` being applied to the previous posting: I don't think it's part of Beancount's contract to guarantee parsed posting ordering. There are some challenges related to booking that we may end up by sorting postings (see https://beancount.github.io/docs/beancount_v3.html#self-reductions).  However this is not an issue as long as we can return the postings to file order using the lineno meta.

I'm coming around to option 2, perhaps it's the best option after all. We can do it in `group_postings_by_meta` to keep the sources consistent.  If we're all agreed I can implement that.

Zburatorul:
> Good points. I wish that Schwab didn't remove their OFX endpoint, or that they'd let us build against their real APIs. I applied as a dev in Nov 2020 after they killed their OFX gateway.

So have I. They have been surprisingly quiet about this.

carljm:
> Could the schwab source be changed to not use the amount at all for matching against imported data?

I don't like the idea that a posting could match even with wildly different amounts; that feels like a significant step back in the rigor of the matching.

> I'm coming around to option 2, perhaps it's the best option after all. We can do it in group_postings_by_meta to keep the sources consistent. If we're all agreed I can implement that.

That works for me! Thanks for your work on this.

> I don't think it's part of Beancount's contract to guarantee parsed posting ordering

If we aren't duplicating the meta, then I don't see a great alternative to the "merge with previous posting" approach; doing this by lineno seems workable.

> I wish that Schwab didn't remove their OFX endpoint

Apparently it still exists, they've just locked it down to make sure only Intuit / H&R Block products can access it :/

bayesianmind:
I've implemented this as discussed and some tests for it. Ready for re-review @carljm @jbms

setup.py
```diff
@@ -179,6 +179,9 @@ def run(self):
     test_requirements=[
         'pytest',
         'coverage',
+        'types-setuptools',
+        'types-atomicwrites',
+        'types-python-dateutil'
```
bayesianmind, line 184: NB: adding these here didn't resolve the build error, that was fixed with the change to tox.ini below. So it's possible this is completely wrong.

PR #111, opened 2021-04-25
Support "Stock Split" in Schwab CSV importer

carljm:
This adds support for the "Stock Split" action in a Schwab CSV transaction file.

If the optional lot-details feature is not in use, the generated transaction will likely need manual updating; this is pretty much unavoidable since there isn't enough information in the transactions CSV to generate a transaction that updates all existing lots with the stock split.

If the optional lot-details feature is in use, then the importer will generate a correct transaction that, for each lot, subtracts all current shares at the old price and adds all new shares at the new price, with lot date preserved. The only information in the transaction CSV is the number of shares added, so the stock split ratio is calculated from that as `(existing_shares + added_shares) / existing_shares`.

Includes a unit test for the new `LotsDB.split()` method, and end-to-end tests both with and without lot details.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -150,35 +150,38 @@ from beancount_import.unbook import group_postings_by_meta, unbook_postings
 CASH_CURRENCY="USD"
 
 class BrokerageAction(enum.Enum):
-    CASH_DIVIDEND = "Cash Dividend"
-    PRIOR_YEAR_CASH_DIVIDEND = "Pr Yr Cash Div"
-    SPECIAL_DIVIDEND = "Special Dividend"
-    QUALIFIED_DIVIDEND = "Qualified Dividend"
-    BUY = "Buy"
-    SELL = "Sell"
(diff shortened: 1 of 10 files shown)


carljm:
Would be glad for review by @bayesianmind and @Zburatorul, if you have time!

beancount_import/source/schwab_csv.py
```diff
@@ -598,6 +610,74 @@ def get_narration_prefix(self) -> str:
...
+                        label=None,
+                        merge=None,
+                    ),
+                    price=None,
+                    flag=None,
+                    meta=self.get_meta(),
+                )
+            )
+
+        postings.append(
```
Zburatorul, line 662: Do splits ever involve some forced selling?
If the arithmetic is correct, there should be no need for any other posting.

carljm: You're right, and good catch! Will remove this. I think in practice I had a need for this posting in one split due to inexact cost figures, so a penny or two needed absorbing. But I don't think with this automation that should occur, and it's better to make the error explicit if it does.

bayesianmind (approved):

PR #112, opened 2021-04-25
Fix PayPal test failures.

Zburatorul:
Not sure how this happened but `pytest -k paypal` on master fails.


Zburatorul:
Nevermind, I found a random uncommitted PayPal json file in my tree.

PR #117, opened 2021-06-04
accept OTHERVEST OFX clause

balkjw:
This commit allows the 'OTHERVEST' clause to be imported by beancount-import.

Per the OFX spec, in <INV401KSOURCE>, "Default if not present is OTHERNONVEST. "

I tested this commit against my Vanguard OFX file, and against the builtin beancount-import tests.


Zburatorul:
Cool. Can you include a small sanitized sample OFX file (or modify an existing one) and test journal file so we can test this automatically in the future?

balkjw:
Done. I added a clause to vanguard401k.ofx and the associated journal entry, and the test passed.
One thing I wanted to bring to your attention is that the transaction gets added to the following account, since the source(s) is/are unknown:
   Assets:Retirement:Vanguard:Company401k  7.62519 VANGUARD-92202V351 {44.13 USD}

If this isn't to your liking, it could probably be changed to something like:
 Assets:Retirement:Vanguard:Company401k:OTHER

Zburatorul:
> One thing I wanted to bring to your attention is that the transaction gets added to the following account, since the source(s) is/are unknown

This new transaction type represents a vesting event I take it, which means an asset is transferred into your ownership.
So that destination account makes sense to me. For the source account (which would have to be of Income type), I don't see a lot of value in adding a new meta key to the asset account and teaching the ofx importer to use that income account. I think b-i will learn to assign it to the right income account anyway.

Zburatorul:
Thanks for your contribution @balkjw!

PR #119, opened 2021-09-05
Add more checks for when number is missing

cookie223:
When using `beancount-import` I often encounter errors like 
```
- is not a valid binary operator between type and Decimal
- Amount's number is not a Decimal instance: MISSING
```
I assume that's caused by omitting amounts in my postings. For this PR just adding more checks to skip those positings without throwing errors.


jbms:
Thanks, can you add some test cases that demonstrate the issue?

Is this caused by actual valid postings, or only if the journal is somehow invalid?

cookie223:
The journal should be valid. Beancount itself allows one missing number per currency so something like this is valid:
```
2021-03-12 * "Delta" "SEA - JFK Jun 19"
  Expenses:Travel:Airfare                            48000.00 P_DL
  Assets:Points:DeltaAirlines                                      P_DL
  Liabilities:CreditCard:Amex:BizPlat                 -5.60 USD
  Expenses:Travel:Airfare                                            USD
```

jbms:
While it is true that beancount allows one missing number per currency, beancount-import performs booking prior to matching which should resolve the amounts.  I think there should only be missing amounts if booking fails for some reason.

cookie223:
I went install 1.3.3 again and checked the first error:
```
matching.py", line 197, in get_posting_weight
    return amount_mul(posting.price, posting.units.number)
amount.py", line 158, in mul
    assert isinstance(amount.number, Decimal), (
AssertionError: Amount's number is not a Decimal instance: <class 'beancount.core.number.MISSING'>
```
The transaction in the ledger is:
```
2019-09-01 * "Easyjet" "LGW-GIB EX57SM1"
  Expenses:Travel:Airfare                             63.53 EUR @ 
  Liabilities:CreditCard:USBank:Altitude             -82.95 USD
```
and posting evaluates to
``` (Pdb) posting
Posting(account='Expenses:Travel:Airfare', units=63.53 EUR, cost=None, price=<class 'beancount.core.number.MISSING'> <class 'beancount.core.number.MISSING'>, flag=None, meta={'filename': '/home/qchu6/accounting/2019/travel.bean', 'lineno': 54, '__automatic__': True})
```
However I couldn't reproduce it in the `manually_entered` example. I wonder if the posting is supposed to have correct price after booking?

Zburatorul:
> The transaction in the ledger is:
> 
> ```
> 2019-09-01 * "Easyjet" "LGW-GIB EX57SM1"
>   Expenses:Travel:Airfare                             63.53 EUR @ 
>   Liabilities:CreditCard:USBank:Altitude             -82.95 USD
> ```

I believe this is invalid syntax. See: https://beancount.github.io/docs/beancount_language_syntax.html#costs-and-prices
You have to have a price if you're going to have a single @.
Does this resolve your original errors, or are those unrelated?

cookie223:
> > The transaction in the ledger is:
> > ```
> > 2019-09-01 * "Easyjet" "LGW-GIB EX57SM1"
> >   Expenses:Travel:Airfare                             63.53 EUR @ 
> >   Liabilities:CreditCard:USBank:Altitude             -82.95 USD
> > ```
> 
> I believe this is invalid syntax. See: https://beancount.github.io/docs/beancount_language_syntax.html#costs-and-prices You have to have a price if you're going to have a single @. Does this resolve your original errors, or are those unrelated?

It's a simplified version of ` 63.53 EUR @@ 82.95 USD`. I like to have 82.95 USD on the line of where the actual transaction happened so I have this syntax in many places.

cookie223:
any updates on this? Shame to have to modify my local libs everytime I upgrade beancount-import.
The `@` or `{}` at the end syntax is valid and works with fava as well.

jbms:
Sorry for the long delay.  If you can modify this pull request to also include a test case that fails without this change, I think we are good to merge.

jbms:
I see you added a journal editor test, but you didn't add a matching test.  The journal editor test does not actually exercise any of the code in `matching.py`.

cookie223:
The base branch is too old. will try again in #149

PR #120, opened 2021-09-10
add some more actions

cookie223:
also had problem with schwab_to_account map, some entries have more than just XXXX-1234 part.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -164,6 +164,7 @@ class BrokerageAction(enum.Enum):
     JOURNALED_SHARES = "Journaled Shares"
     LONG_TERM_CAP_GAIN = "Long Term Cap Gain"
     MARGIN_INTEREST = "Margin Interest"
+    CREDIT_INTEREST = "Credit Interest"
     MISC_CASH_ENTRY = "Misc Cash Entry"
     MONEYLINK_TRANSFER = "MoneyLink Transfer"
     PRIOR_YEAR_CASH_DIVIDEND = "Pr Yr Cash Div"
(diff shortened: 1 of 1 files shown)


beancount_import/source/schwab_csv.py
```diff
@@ -1098,7 +1104,7 @@ def __init__(self, journal: JournalEditor, lots: LotsDB) -> None:
         self.found_accounts: Set[str] = set()
 
     def process_entry(self, raw_entry: RawEntry) -> Optional[TransactionEntry]:
-        account = self.schwab_to_account.get(raw_entry.account)
+        account = self.schwab_to_account.get(raw_entry.account) or self.schwab_to_account.get(raw_entry.account[-9:]) # try only the XXXX-1234 part
```
carljm, line 1107: Do you know for sure that this "initial part" is always exactly 9 characters long? What is the "initial part" in your case? Can we have a more robust way to identify the real account number and strip off the rest? Ideally where we construct the account map, not here...

carljm (changes requested):
New actions look good, would be happy to merge those. The way of handling the account map I'm  not sure about.

beancount_import/source/schwab_csv.py
cookie223: I think I did it here because in some files it is "XXXX-1234" (length 9), in some files it is like "IRA XXX-1234". We can leave this part to another time.

carljm: I filed #122 for follow-up on this.

Zburatorul:
@cookie223, can you add some lines with these new transactions types to the test files under `testdata/source/schwab_csv/test_basic/`. It sounds like the rest of us don't have these action types so can't verify what the expected inputs and outputs should be.

carljm:
Yes, it would be ideal to add some tests for these; if you can provide sample lines here I can also update the tests with them. Thanks @Zburatorul for the reminder.

Issue #122, opened 2021-09-15
Handle variations in Schwab account name format in accounts map


carljm:
See changes removed from #120

cookie223:
The positions csv file I got has "Positions for account Individual XXXX-1234 as of 08:42 PM ET, 08/22/2021" while the transaction csv I got has "Transactions for account XXXX-1234 as of 08/22/2021 20:42:49 ET from 07/01/2021 to 08/21/2021".
Basically we need a way to match "Individual XXXX-1234" and "XXXX-1234" to the same account. Or maybe when parsing only take out the XXXX-1234 part.

Zburatorul:
I think we should change the `TITLE_RE` regex. I propose we do it like this:
https://regex101.com/r/DBoe1R/1

This matches all possible variations I have heard of in the discussion and does not make any new assumptions imo.

carljm:
Yes I think that looks good, will submit a PR with that change and associated tests.

carljm:
#123 should fix this. I didn't use your regex exactly, @Zburatorul, because positions title regex and transactions title regex still need to be a bit different due to needing to extract the date, which is in slightly different formats. But I did extract a common "account name regex" that both use, which uses your approach to extract only the `XXXX-1234` part. Your review welcome!

carljm:
Fixed with merge of #123

PR #123, opened 2021-09-15
[schwab] trim account IDs for consistency across CSVs

carljm:
Previously we read the full account name from CSVs (e.g. `Brokerage XXXX-1234`) and used that as the account identifier. I'd previously observed that Lot details CSVs didn't use that full account name, just the `XXXX-1234` part, and so I'd added an ugly hack in the LotsDB code (the `convert_account` method) so the account names would match up. It now turns out (see #122) that sometimes other CSVs also don't use the full name. So this PR removes the hack from `LotsDB` and instead across the board uses only `XXXX-1234` account identifiers, stripping off the text before that when reading the accounts from CSV.

For backward compatibility I don't want to require people to update the `schwab_account` metadata on their Beancount accounts for things to keep working, so when we read that metadata from account meta we also trim it to just `XXXX-1234` if it says e.g. `Something XXXX-1234`.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -280,6 +280,7 @@ class RawBrokerageEntry(RawEntry):
         interest_account = self.get_meta_account(account_meta, INTEREST_INCOME_ACCOUNT_KEY)
         dividend_account = self.get_meta_account(account_meta, DIV_INCOME_ACCOUNT_KEY)
         taxes_account = self.get_meta_account(account_meta, TAXES_ACCOUNT_KEY)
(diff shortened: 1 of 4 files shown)


Zburatorul:
I looked over the changes and they look good to me.

Zburatorul (approved):

Zburatorul:
I noticed we have a bunch of `get_schwab_account_from_meta` calls all in `get_processed_entry`.
Let's have a single one at the beginning of the method, just like the regular `get_meta_account` calls. It will also be more clear that they are on equal footing, but the base account one has to handle a special case.

PR #124, opened 2021-09-16
add lines of entries for new brokerage actions

cookie223:
Using github.com editor in browser not sure what's changed for L1-L8.
Adding sample lines for new actions added in #120

PR #125, opened 2021-09-19
[schwab] add tests for new action types

carljm:
Follow up from #120, replaces #124


testdata/source/schwab_csv/test_basic/import_results.beancount
```diff
@@ -724,8 +756,29 @@
...
+;               },
+;               "source_account": "Assets:Schwab:Brokerage-1234:Cash"
+;             }
+;           ]
+2021-01-30 * "MARGININTEREST - SCHWAB1 INT 12/30-01/29"
+  Assets:Schwab:Brokerage-1234:Cash   0.02 USD
+    date: 2021-01-30
+    schwab_action: "Credit Interest"
+    source_desc: "SCHWAB1 INT 12/30-01/29"
+  Expenses:FIXME                     -0.02 USD
```
Zburatorul, line 778: This should be the `interest_income_account`

testdata/source/schwab_csv/test_basic/import_results.beancount
```diff
@@ -811,3 +864,24 @@
...
+;               },
+;               "source_account": "Assets:Schwab:Checking"
+;             }
+;           ]
+2021-07-30 * "ATMREBATE - ATM Fee Rebate"
+  Assets:Schwab:Checking   3.00 USD
+    date: 2021-07-30
+    schwab_action: "ATMREBATE"
+    source_desc: "ATM Fee Rebate"
+  Expenses:FIXME          -3.00 USD
```
Zburatorul, line 887: This should be the `fees_account`.

Zburatorul (changes requested):
There are these two places where we should be able to identify the right account to use based on the meta keys. I missed that in the original PR somehow.
If we leave the FIXME, b-i should learn the right account over time, but seems like a simple fix.

carljm:
@Zburatorul Great catch! Fixed these, as well as a couple other misc things I noticed.

carljm:
Build breakage here is from breakage in main branch; fixed separately in #127

testdata/source/schwab_csv/test_basic/import_results.beancount
```diff
@@ -1,23 +1,13 @@
...
-;           ]
-2018-06-28 * "MARGININTEREST - INTEREST 05/30THRU 06/27"
+; features: []
+2018-06-28 * "Margin Interest - INTEREST 05/30THRU 06/27"
   Assets:Schwab:Intelligent-4321:Cash  -0.05 USD
     date: 2018-06-28
     schwab_action: "Margin Interest"
     source_desc: "INTEREST 05/30THRU 06/27"
-  Expenses:FIXME                        0.05 USD
+  Income:Interest:Schwab                0.05 USD
```
Zburatorul, line 10: I can see that the right account is filled in here, but it's strange that the features array is now empty. Is this a regression?

testdata/source/schwab_csv/test_basic/import_results.beancount
```diff
@@ -811,3 +844,14 @@
     date: 2021-03-25
     schwab_action: "Stock Merger"
     source_desc: "SPAC MERGER CORP XXXMANDATORY MERGER EFF: 03/25/21"
+
+;; date: 2021-07-30
+;; info: {"filename": "<testdata>/test_basic/transactions/EAC_Checking_Transactions_20201223-160009.CSV", "line": 5, "type": "text/csv"}
+
+; features: []
```
Zburatorul, line 851: Should `features` be empty?

Zburatorul (commented):
Looks good overall. I am unclear about whether the `features` arrays in the test output files should always be non-empty?

carljm:
I believe it is expected for features to be empty if there is no fixme posting in the txn. Features are used for the Bayesian prediction of fixme accounts.

Zburatorul:
> I believe it is expected for features to be empty if there is no fixme posting in the txn. Features are used for the Bayesian prediction of fixme accounts.

I see. In that case it's good to merge. Thank you.

PR #127, opened 2021-09-26
Fix mypy and tests.

carljm:
Both type checking and the Paypal tests were broken by recent changes. This makes the build green again. For the paypal tests I just allowed them to rewrite the expected output, so a reviewer should verify that the output changes are in fact correct.


jbms:
I apologize for introducing those failures.

I am working on a fix now for paypal.  Unfortunately it appears that Paypal has changed how it treats transactions.

jbms:
Okay, I just pushed a fix to the build breakage.  Next time I'll be sure to use a pull request for my own changes as well...

carljm:
No worries, thanks for the fixes!

PR #128, opened 2021-10-01
Allow sources to skip accounts for training examples

m-d-brown:
This is more explicit and general than
training.DEFAULT_IGNORE_ACCOUNT_FOR_CLASSIFICATION_PATTERN. For my case
I call my capital gains account 'Income:CapitalGains' instead of
'Income:Capital-Gains' so the pattern did not apply. I was initially puzzled
why my transacations were not being used as training examples.

Hopefully you don't find this new metadata to be too much overhead.


jbms:
It looks like there is already an option, `ignore_account_for_classification_pattern`, that gets partially passed through.  However, I see that there is no command-line argument `--ignore-account-for-classification-pattern` to actually set the option.

In general I think setting a pattern like this would be preferable to adding an extra metadata key to every ofx transaction with capital gains.  Alternatively perhaps there could be some mechanism for the ofx source to indicate that the capital gains accounts should not be used for training, without actually adding metadata.  E.g. sources could be allowed to specify additional training-exclusion accounts when they are initialized.

m-d-brown:
Not having extra metadata makes sense. I've looked more at the implementation and see now that a source can make the not-for-training accounts available during initialization. Much simpler.

I've tried this with my latest commits, using SourceResults since the authoritative accounts are already passed there. I also added some structure in OFX to show the fees and gains accounts aren't the primary accounts, to make their role easier to see. It turned out to be more code but hopefully worthwhile.

beancount_import/training.py
```diff
@@ -126,8 +130,22 @@ def default_extractor(x, key_value_pairs, key=key):
...
 
     def _ignore_posting_for_automatic_classification(self, posting):
-        return re.match(self.ignore_account_pattern,
+        skip = self.skip_accounts_cache.get(posting.account)
+        if skip is not None:
+            return skip
+
+        # Cache because checking prefixes can be an expensive O(m*n)
+        # operation.
+        for parent in self.skip_accounts:
```
jbms, line 139: I like the cache.  You could additionally possibly make the matching more efficient by dynamically constructing a regular expression from the set.  But I think it is fine as is.

m-d-brown: I like that idea. Done. I've also switched to functools.lru_cache to make this simpler.

beancount_import/training.py
```diff
@@ -104,11 +106,15 @@ def __init__(
             self,
             account_source_map: Dict[str, 'Source'],
             sources: Iterable['Source'],
+            skip_accounts: Set[str],
             ignore_account_pattern:
             str = DEFAULT_IGNORE_ACCOUNT_FOR_CLASSIFICATION_PATTERN,
     ) -> None:
         self.account_source_map = account_source_map
-        self.ignore_account_pattern = ignore_account_pattern
+        self.skip_accounts_re = [re.compile(p) for p in
```
jbms, line 114: You can combine it into a single regular expression (using `(?: )` and `|`), which should be much more efficient to evaluate.

m-d-brown: Okay, that's better and simpler. Done.

PR #131, opened 2021-10-10
generic_importer_source: apply costs when balancing

jakobwenzel:
The Generic Importer Source did not apply Costs at all when balancing entries. This PR tries to fix that.


jbms:
Thanks.

Can you add a simple test case?

jakobwenzel:
I added a testcase. The existing testcases for the generic source use `CSVImporter`, which cannot handle costs. Therefore, my testcase loads the transaction from a beancount file.

jbms:
Thanks, please revert the package.json lockfile changes though.

jakobwenzel:
Oops, that's what I get from using `git commit -a`...

jbms:
Thanks!

PR #133, opened 2021-11-11
Improve ultipro_google import: regression test, efficiency improvement, and employer contribution matching

m-d-brown:
For the employer contribution matching I assume all callers would prefer to have the corresponding postings appear. If not I could make this configurable. I was using my own importer previously but yours is much better, except that it does not do matching.


jbms:
Thanks.

Looks like we need to modify the github actions config to ensure `pdftotext` is available.

beancount_import/source/ultipro_google_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+testdata_dir = os.path.realpath(
+    os.path.join(
+        os.path.dirname(__file__), '..', '..', 'testdata', 'source', 'ultipro_google'))
+
+examples = [
+    ('test_basic', '20210108-1.pdf'),
+]
+
+@pytest.mark.parametrize('name,ofx_filename', examples)
```
jbms, line 16: Should be pdf_filename rather than ofx_filename

m-d-brown: This made me realize that the filename isn't used at all. Removed.

m-d-brown:
pdftotext is through poppler so I don't think there's a good way to make it available in the different environments run on Actions. I'll think about it more. One approach may be to depend on https://pypi.org/project/pdftotext/ rather than shelling out to an external binary.

jbms:
On ubuntu you can install with apt-get.  The test could be skipped on other platforms.

m-d-brown:
Updated to install poppler-utils for Ubuntu and run the test only if pdftotext is available. I've also made it skip Windows because the parser doesn't handle Windows newlines correctly.

Build is already failing on the master branch for macos, however, after d837556 (@jakobwenzel, FYI).

m-d-brown:
The generic importer fix is pretty easy. #134 should fix it.

PR #134, opened 2021-11-13
Simplify and better encapsulate generic_importer test_cost.

m-d-brown:
Since 'beancount' files can't be identifed, use 'txt' instead.  Commit 1cedf933c8f50e773b51287f4465f2a51f24bc66 earlier fixed similarly broken tests. This fixes the breakage after #131.


carljm:
@m-d-brown I merged #136, which gives the generic importer the ability to recognize beancount files without needing libmagic. With that in place so the tests can work with a `.beancount` file, are there other changes in this PR that still have value to merge?

m-d-brown:
What's remaining simplifies test_cost so I think it's still worthwhile to merge, though I wouldn't be too bothered to drop it.

carljm:
Looks good, thanks!

PR #135, opened 2021-11-14
[schwab_csv] add Cash In Lieu action

carljm:

PR #136, opened 2021-11-14
Fix test failures on Windows (or anywhere libmagic is not installed.)

carljm:
#131 reintroduced a case of the generic-importer tests trying to guess the mimetype of a beancount file, which fails if libmagic is not installed. This is now showing up on new PRs (eg #135 ) as failures of the Windows tests, which don't have libmagic; not sure why this didn't get caught on the PR itself.

This fixes it by explicitly adding support to the generic importer to recognize beancount files as `text/plain`, which matches what libmagic guesses anyway.

### DIFF
--- a/beancount_import/source/generic_importer_source.py
+++ b/beancount_import/source/generic_importer_source.py
@@ -58,7 +58,7 @@ class ImporterSource(DescriptionBasedSource):
 
     def prepare(self, journal: 'JournalEditor', results: SourceResults) -> None:
         results.add_account(self.account)
-        
+
         entries = OrderedDict() #type: Dict[Hashable, List[Directive]]
         for f in self.files:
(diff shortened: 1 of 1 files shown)


m-d-brown:
My #134 fixes the breakage after #131 as well, in a different way.

carljm:
> My #134 fixes the breakage after #131 as well, in a different way.

Oh, I missed that, thanks for pointing it out! I think that's a great approach too but I would suggest whether we merge that or not, either way we should merge this one. This issue (of the generic importer trying to detect mimetype of beancount files) has crept into the generic importer tests twice now, and I think it's reasonable to assume it'll do so again; this fix makes that a non-issue.

Zburatorul (approved):
LGTM

PR #137, opened 2021-11-14
A merged posting is implicitly cleared.

carljm:
#110 introduced `merge: TRUE` metadata which can be used on manually booked lots from import sources that don't have unique transaction IDs, so that unbook can re-merge them and they'll match the un-split imported posting.

A posting with `merge: TRUE` should also be implicitly considered cleared, because it is really part of the posting above that it merges with. If the merged-with posting is uncleared, it will show up as such and we don't need both postings showing up as uncleared. If the merged-with posting is cleared, then the merged one is too.

Without this we have to manually add `cleared: TRUE` to every posting with `merge: TRUE` once its merged-with posting has cleared.

### DIFF
--- a/beancount_import/source/description_based_source.py
+++ b/beancount_import/source/description_based_source.py
@@ -129,4 +129,6 @@ class DescriptionBasedSource(Source):
     def is_posting_cleared(self, posting: Posting):
         if posting.meta is None:
             return False
+        if posting.meta.get("merge"):
+            return True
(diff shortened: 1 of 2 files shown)


carljm:
Windows tests need #136

carljm:
Also cc @bayesianmind .

bayesianmind (approved):
Makes sense, thanks!

PR #138, opened 2021-11-14
[schwab_csv] resolve lot asof to nearest day.

carljm:
The auto-lot-split logic relies on grouping all lot details downloaded at the same time and assuming they are complete for that point in time; any missing lot-details CSV for a commodity will result in an assumption that zero units of that commodity were held at that point in time.

But each lot details CSV will be downloaded a second or two after the one before it, even when using finance-dl to automate the download, and this could cross a minute boundary. Then we will have some lot details downloaded at minute X and some at minute X+1; these will be treated as downloaded at a different point in time, splitting what should have been one point-in-time into two, each with many missing commodities.

This PR resolves this by truncating the asof-datetime to just the date, so all lot details downloaded on the same date will be considered the same point in time.

The original problem could still occur if lot details are downloaded at exactly midnight, but this seems unlikely and I don't see a way to avoid that.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -444,7 +444,7 @@ class RawLot:
     """A single cost-basis lot of a single holding from Lot Details CSV."""
     symbol: str
     account: str
-    asof: datetime.datetime
+    asof: datetime.date
     opened: datetime.datetime
     quantity: Decimal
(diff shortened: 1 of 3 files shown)


carljm:
Windows tests need #136

Zburatorul (approved):
LGTM.

PR #139, opened 2021-11-14
[schwab_csv] handle extended lot-details CSV format.

carljm:
When a lot has disallowed loss, Schwab renders the lot-details CSV for that commodity in an extended format. This PR handles that extended format.

In some cases (I'm not sure if related to above or not) the time portion of a lot open date is zeroed out, when in previous lot details a time was included for that lot. Handle this by tracking lot open date simply as a date, not as a datetime.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -445,7 +445,7 @@ class RawLot:
     symbol: str
     account: str
     asof: datetime.date
-    opened: datetime.datetime
+    opened: datetime.date
     quantity: Decimal
     price: Decimal
     cost: Decimal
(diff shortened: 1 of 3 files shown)


carljm:
Windows tests need #136

Zburatorul (commented):
Let me make sure I understand the logic.
Afaik, disallowed loss transactions are examples where the wash-sale rule applies. If such a transaction is followed up by buying the same security within 30 days, the cost basis in the second transaction would be higher by the disallowed loss amount.
The way we currently book things, that will not happen. The ledger will reflect a loss at sale time (first transaction), and a cost basis that's too low in the second transaction.
Fixing this would require the importer to identify old transactions that qualify as wash-sale and transfer the loss as a cost-basis adjustment.

@carljm, at the moment we're just assuming all of that is done manually, right?

carljm:
@Zburatorul correct, I haven't attempted yet to actually automate correcting the cost basis for disallowed loss. The goal of this PR is just to permit import with the extended lots CSV format that occurs when there is a disallowed loss, so that import isn't totally blocked on this scenario.

Zburatorul (approved):

PR #140, opened 2021-11-14
[schwab_csv] add support for reverse split.

carljm:


carljm:
Windows tests need #136

Zburatorul (approved):
Looks sensible. It seems that mergers, splits and reverse splits should be all be handled by some unified logic, but that definitely warrants another PR.

PR #142, opened 2021-12-11
[Paypal] Add locales option, support for Paypal DE

moritzj29:
Hello,
first of all thanks for this great codebase and this great tool! I'm just starting out using beancount and beancount-import and I'm pretty excited. To get all my data imported correctly, I added functionality to enable import for Paypal DE (Germany) transactions. 

This requires two changes:

- parsing of the date string in German rather than in English
- take care of the different use of `,` and `.` as number separators in Germany compared to the rest of the world

Therefore I added a `locale='DE'` option to make this easily extendable, if required in the future. `locale='EN'` is set as default, so the code should be backwards compatible.

I'm happy for your feedback! So far I did not find any example on how to handle locales neither for beancount nor for beancount-import, so I hope this approach is fine with you.

**ToDo's**
- [x] add some tests
- [x] resolve if `fundingSource` is always set in `DE` for `SEND_MONEY_RECEIVED` (#143), eventually add handling to locales
- [x] add `earliest_date` option
- [x] update `locale` names according to @jbms remarks in #144


moritzj29:
so I cleaned up the commit history and factored out the more groundbreaking changes in #162 and #163.
The code, especially the tests, require those PRs to be merged first. But I think it is better to keep them separated since they are not directly related to the locales functionality added in this PR.

I'm happy for your review and your thoughts!

Zburatorul:
Looks good overall. Can you figure out why it's failing for python3.9?

rkok:
Hi @moritzj29, first off, thanks for your work and it indeed looks like you've thought this out a lot more than I did in #161!

I have a concern about the practicality of your solution however: it would require anyone from a non-US/DE locale to create a Pull Request adding their custom locale with all the (likely duplicate) formatting/parsing rules.

What's your take on the following?:

1. Always strip out Unicode "\xa0" regardless of the locale
2. Add an argument to `PaypalSource` for specifying the thousands separator and the decimal separator. (Default: US format for backwards compatibility)
3. Find a more generic way to handle internationalized dates. (?) Or...
    (alt.) Add an argument to `PaypalSource` to pass a user-defined locale class

Happy to hear your take on this as well @Zburatorul

Zburatorul:
> it would require anyone from a non-US/DE locale to create a Pull Request adding their custom locale with all the (likely duplicate) formatting/parsing rules.

Agreed, it would.

>Add an argument to PaypalSource for specifying the thousands separator and the decimal separator. (Default: US format for backwards compatibility)

This solves part of the problem, but parsing the date in de_DE requires more.

>Find a more generic way to handle internationalized dates. (?) Or...
(alt.) Add an argument to PaypalSource to pass a user-defined locale class

Which I agree does not require a PR, but does require writing some code. 

Given that de_DE is the first locale PR/issue we've had for the PayPal source, it seems that either solution is fine and won't produce a lot of frustration in absolute terms. I personally lean towards us managing per-locale classes in `paypal.py` but having the documentation make it clear which locales are supported and that it is easy to add another one.

moritzj29:
great that some discussion is going on here! I really appreciate the interest and the feedback on the PR!

Regarding @rkok remarks, I agree with @Zburatorul 's opinion.

To give a little background: I also work on adapting the amazon importer for DE/multi-language and there way more language-specific customizations are required since the complete invoice needs to be translated. Paypal is quite easy to transform since the JSON is basically the same for both languages, except for date and number separators.
Anyhow, I chose a consistent approach using locale specific classes since they can contain methods as well (not only constants).

I absolutely favor the approach of contributing language specific code back to the codebase (or at least maintained forks). Of course it is easy to say "just pass an appropriate locale class" to the importer, but nobody will contribute this code back to the community. Everybody will start from scratch and it may be quite some work 😄 (check out the Amazon PR). Of course it requires more work of the maintainers, since the codebase grows. But at the same time hopefully the number of users and contributors grows.

moritzj29:
> Looks good overall. Can you figure out why it's failing for python3.9?

Well, after some hours of researching on typing and `mypy` I did 🤣 But since I have basically no experience with this, please check if the chosen approach of defining a base metaclass is a good way...

> Locale should be `en_US` because `en_EN` is apparently not ISO compliant.
Agreed. I changed it.

I also decided to merge #162 and #163 since otherwise the CI would fail because of the failing tests.

Zburatorul:
> > Looks good overall. Can you figure out why it's failing for python3.9?
> 
> Well, after some hours of researching on typing and `mypy` I did rofl But since I have basically no experience with this, please check if the chosen approach of defining a base metaclass is a good way...

Thanks for getting to the bottom of this! And thank you for the continued contributions!

PR #144, opened 2021-12-12
[Amazon] Amazon import source for amazon.de

moritzj29:
Hello,

this is a proposal to generalize the amazon import source to allow additional languages. I added support for German files from amazon.de. For other languages it will be required to read out all the language specific strings for the regexes.

The code is working already but probably needs some polishing.

I added lots of debug logging output, I hope this fine. Otherwise you are completely lost finding the translated regexes...

It runs fine on the tests available, so it should be backward compatible.

Sorry that the code diff is so bad! I moved almost all existing methods into a new class, therefore the changed indentation results in every line marked as changed...

I mark this as draft for now, but comments are very welcome!

- [x] add tests for amazon DE
- [x] same extent of logging in all methods, remove duplicates
- [x] add `locale` argument for direct calling
- [x] rename `locale` to `de_DE` etc.
- [x] minimize diff volume by moving methods outside the class, just pass `locale` as argument


testdata/source/amazon/D56-5204779-4181560.json
```diff
@@ -54,7 +54,7 @@
         }
     ],
     "pretax_adjustments": [],
-    "tax": [],
+    "tax": null,
```
moritzj29, line 57: I changed this, since otherwise the code (also the original?!) throws an exception here:
https://github.com/jbms/beancount-import/blob/dca340694ac60dd8ee7529cc4740b09fe729dbc2/beancount_import/source/amazon.py#L390

so I set `invoice.tax=None` instead of `=[]` for digital invoices

jbms:
Perhaps you can keep the parsing functions as free functions, but pass in the locale object as a parameter --- that way the diff would be much smaller?

jbms:
Also, locales are commonly named like "en_US" or "de_DE".  You are just using the language, but given that you are including such things as whether the tax is included, it should probably be a full locale.

moritzj29:
very valid points, I really appreciate the feedback! I will change the code accordingly!

testdata/source/amazon/D56-5204779-4181560.json
Zburatorul: Can we change the code instead, so all the fields in the json follow the same convention (`field = []`).

moritzj29: makes sense, I changed the code in `amazon.py` to handle `[]`

beancount_import/amount_parsing.py
```diff
@@ -30,12 +30,18 @@ def parse_amount(x, assumed_currency=None):
     if not x:
         return None
     sign, amount_str = parse_possible_negative(x)
-    m = re.fullmatch(r'(?:[(][^)]+[)])?\s*([\$€£])?((?:[0-9](?:,?[0-9])*|(?=\.))(?:\.[0-9]+)?)(?:\s+([A-Z]{3}))?', amount_str)
+    m = re.fullmatch(r'(?:[(][^)]+[)])?\s*([\$€£]|[A-Z]{3})?\s*((?:[0-9](?:,?[0-9])*|(?=\.))(?:\.[0-9]+)?)(?:\s+([A-Z]{3}))?', amount_str)
```
moritzj29, line 33: allow to specify currency with prepended three letters, e.g. `EUR 20.00`

beancount_import/source/amazon_invoice.py
```diff
@@ -15,6 +17,256 @@
...
+    shipment_sales_tax='Sales Tax:'
+    shipment_total='Total for This Shipment:'
+    shipment_seller_profile=' (seller profile)'
+    shipment_sold_by=r'(?P<description>.*)\n\s*(?:Sold|Provided) by:? (?P<sold_by>[^\n]+)'
+    shipment_condition=r'\n.*\n\s*Condition: (?P<condition>[^\n]+)'
+    regular_total_order='Grand Total:'
+    regular_estimated_tax = 'Estimated tax to be collected:'
+    regular_order_placed=r'(?:Subscribe and Save )?Order Placed:\s+([^\s]+ \d+, \d{4})'
+    regular_order_id=r'.*Order ([0-9\-]+)'
+    gift_card='Gift Cards' # not confirmed yet!
```
moritzj29, line 145: I don't know how gift cards are handled on `.com`, these are just guesses inferred from `.de`. There (electronic) giftcards are handled differently than regular shipments. They have an own "shipment table".

Everything is handled in the new `parse_gift_cards` method.
If there is no table node with text matching`gift_card`, the code doesn't do anything. (probably the case for `en_US`) I decided not to add an additional boolean flag to enable/disable the `parse_gift_cards` method.

beancount_import/source/amazon_invoice.py
```diff
@@ -15,6 +17,256 @@
...
+    tax_included_in_price=True  # no separate tax transactions
+    payee='Amazon.de'
+
+    # common fields regular and digital orders
+    items_ordered='Bestellte Artikel|Erhalten|Versendet|Amazon-Konto erfolgreich aufgeladen' # Erhalten|Versendet for gift cards
+    price='Preis|Betrag'
+    currency='EUR'
+    items_subtotal='Zwischensumme:'
+    total_before_tax='Summe ohne MwSt.:'
+    # most of translations still missing ...
```
moritzj29, line 181: unfortunately, there is no way to get all these translations (if they exist at all). I parsed all my amazon transactions from the last years and added every case that occurred. Probably there will be minor PRs in the future adding additional translations...

moritzj29:
Sooo, I think this code is finally ready for review now. I tried to keep the changes to a minimum to reduce the diff, but unfortunately I had to cover some additional cases (special gift cards order table) which resulted in splitting some existing methods.

As said in the above code comments, translations (both ways!) may not be complete yet! So minor PRs may follow along the way as we discover additional edge cases...

I ran the code against all amazon transactions I had available (DE and tests) and there were no exceptions thrown. For the last months I thoroughly checked the parsed invoices and the created transactions against my account transactions as well.

I added test cases for every invoice scenario I experienced to ensure code compatibility in the future. I think this makes sense when working with multiple languages since otherwise the contributors will only have their own data at hand for testing.

As always, I'm thankful for any comments and improvements. I really appreciate the effort you guys take to maintain this (and the `finance-dl`) codebase!

beancount_import/source/amazon_invoice_test.py
```diff
@@ -17,7 +17,7 @@
     '166-7926740-5141621',
     'D56-5204779-4181560',
 ])
-def test_parsing(name: str):
+def test_parsing_en_EN(name: str):
```
Zburatorul, line 20: en_US

beancount_import/source/amazon.py
```diff
@@ -387,7 +395,7 @@ def make_amazon_transaction(
                         (INVOICE_DESCRIPTION, adjustment.description),
                     ]),
                 ))
-        if invoice.tax is not None and invoice.tax.number != ZERO:
+        if len(invoice.tax)>0 and invoice.tax.number != ZERO:
```
Zburatorul, line 398: What guarantees invoice.tax is not None?

beancount_import/source/amazon_invoice_test.py
moritzj29: missed this.

beancount_import/source/amazon.py
moritzj29: it relates to the above discussion: https://github.com/jbms/beancount-import/pull/144#discussion_r771871191

But I cannot recall why I removed the check on `None`...

For digital orders, `tax=[]` on order level anyway, so no need to check for `None`.

For regular orders, tax is set via:

```
tax = locale.parse_amount(
        get_field_in_table(payment_table, locale.regular_estimated_tax))
```
Which may return `None` if the field is not found...

Following up on the discussion above, I would vote for checking on `None` in `parse_regular_order_invoice` and setting `tax=[]` in this case. This ensures that in the resulting JSON `"tax": []`. Maybe also logging an error or warning might make sense, since so far all regular invoices contain tax information.

What are your thoughts on this? Did I miss something?

beancount_import/source/amazon_invoice.py
Zburatorul: So what happens if a missing translated string is needed? Presumably an exception because the base class is abstract.

Do you see a way for us to handle this gracefully?
We'd ideally notify the user of the fact that we cannot parse their invoice, not crash, and even suggest they open an issue.

beancount_import/source/amazon_invoice.py
Zburatorul: Here I also think we'd better not guess the correct string to match, but fail gracefully.

Zburatorul (commented):
Thank you for this massive contribution @moritzj29. Great work!
I don't have any objections to the code structure, but I'd like to discuss the user experience a bit.

Right now I see two kinds of undesirable failure modes:
1. The gift card (purchase) matching string is, as you say, a guess. If we're wrong it won't match silently(?) and the user will wonder why the importer didn't add a new transaction.
2. Some missing strings in the de_DE locale. If they are ever needed, the failure will not be gentle. 

I submit that a good UX in both cases is for the importer 1) not to fail 2) not to offer/identify any transactions 3) warn the user that it cannot handle these cases and suggest submitting an issue.
Does anyone see a better way?

In terms of implementation, my impromptu suggestion is to 1) leave the gift card related string empty 2) catch whatever exception the abstract base class throws when importer tries to use one of those fields 3) print a nice message for the user ideally specifying what field name was missing.

moritzj29:
Thank you so much for your comments and working through all that code!

I agree with you that the UX should be as good as possible and crashing of the importer must be avoided.

I will check the impact of the missing translations. Most of them are just additional list entries, so the class attribute itself is available. It just matches less cases.

I think for the `pretax_adjustment_fields_pattern` this should not be an issue. Just the totals will not match and raise an error in the UI. I also guess that most of them don't even exist for DE...

The `parse_gift_cards` methods looks for an appropriate table and does nothing if the table is not found.

I mean, the question boils down to "how to check for elements to be present we don't know yet how to check on". **I think the most robust way would be to raise and catch exceptions. Then the importer would fail gracefully on invoice level.**

As you probably recognized, I added a lot of debug logging statements. Since I found it really difficult to get a grasp on where the code failed when I started porting it. This way one should at least get an idea on where it fails...

Also it would be great to add more testdata, also for en_US, e.g. a corresponding test for all these regexes. I will try to add a README or similar with some images indicating which part of the HTML invoice translate into which pattern. This will make it more easy to compare invoice and code and check for missing cases.

beancount_import/source/amazon.py
moritzj29: So I tried to tackle this and stumbled upon an inconsistency in the current codebase:

According to the typedefs `tax: Amount` for `Order` and `Shipment`. But this is not correct. It is actually `tax: Sequence[Adjustment]` for `Shipment`. So to have an empty `tax` field on shipment level one needs to set `tax=[]`, whereas on order level `tax=None`, since `Amount` can be None.

I don't know why the typedef check does not catch this...

When it comes to digital orders, the current code sets `tax=[]` on order level. This is not correct, since `Amount` cannot be `[]`, it must be `None`. The code works nevertheless, since tax is always given on shipment level for digital orders and therefore tax on order level is not used anyway.

From the code perspective, I would go with the above corrections. But my dilemma arises when it comes to the JSON output for the tests: Setting `tax=None` translates into `tax: null` in the JSON. You suggested above that we should stick to the convention `field=[]`. I would drop this requirement since it would require additional modifications of the data (only for the test files...).

@Zburatorul are you fine with this? what are your thoughts?

moritzj29: I added a proposal in [4be65b3](https://github.com/jbms/beancount-import/pull/144/commits/4be65b33d717c08cf13309efc51dc36c29066f8c)
- fix `shipment.tax` to type `List[Adjustment]`
- fix `order.tax` to type `Amount`
- results in `tax: null` in JSON for orders with no tax transaction (tax not given or included in item price)

beancount_import/source/amazon_invoice.py
moritzj29: if it is not found, it does not even fail. But I made the attribute optional now.

I also added a check if there are no items found at all for the regular order or digital order (no shipments, no gift cards), this should indicate that there is something wrong with the parsing...

beancount_import/source/amazon_invoice.py
moritzj29: exceptions should not happen. The base class is abstract but all attributes are (and have to be!) set in the derived class. Only alternatives may be missing which will not lead to exceptions, but just missed cases.

moritzj29:
I went through the code once again and added missing types, comments and error messages. I think I addressed all of your comments above (see my comments above), at least with a proposal.

I think the code is quite robust now. There are lots of debugging statements and error messages whenever possible. And the whole `amazon_invoice` code is wrapped inside a `try: except:` statement anyway.

I propose that I squash everything into a single commit, once we are done with the review. There have been quite a lot of commits in the meantime...

On the long run, I propose to move the `amazon_invoice` code into its own class. But this should be a separate PR due to the large expected diff...

Zburatorul:
Thanks again @moritzj29.
I squashed the commits myself.

PR #146, opened 2021-12-29
Add further compatibility for Schwab Brokerage Actions

chandler150:
Adds compatibility for Long Term Cap Gain Reinvest, Short Term Cap Gain Reinvest, Bank Transfer, and Moneylink Deposit.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -1,53 +1,41 @@
 """Schwab.com brokerage transaction source.
-
 Imports transactions from Schwab.com brokerage/banking history CSV files.
-
 To use, first you have to download Schwab CSV data into a directory on your filesystem. If
 you have a structure like this:
-
     financial/
       data/
         schwab/
           transactions/
           positions/
-
 And you download your transaction history CSV into `transactions/` and your positions
 statement CSV into `positions/`, then you could specify your beancount-import source like
 this:
-
     dict(module="beancount_import.source.schwab_csv",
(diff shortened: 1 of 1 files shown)


beancount_import/source/schwab_csv.py
```diff
@@ -1,96 +1,75 @@
 """Schwab.com brokerage transaction source.
-
```
carljm, line 2: Please revert the removal of all these blank lines in the docstring.

beancount_import/source/schwab_csv.py
```diff
@@ -327,6 +310,8 @@ def get_processed_entry(
             BrokerageAction.NON_QUALIFIED_DIVIDEND,
             BrokerageAction.QUAL_DIV_REINVEST,
             BrokerageAction.REINVEST_DIVIDEND,
+            BrokerageAction.LONG_TERM_CAP_GAIN_REINVEST,
+            BrokerageAction.SHORT_TERM_CAP_GAIN_REINVEST
```
carljm, line 314: nit: add trailing comma

carljm (changes requested):
Looks good! Please address comments -- seems like your editor is removing all blank lines from docstrings?

Ideally we would add each of these new actions to the test suite to ensure we don't break them in future, but I'm less concerned about that since we are not adding new behavior here, only new aliases for existing behavior.

carljm:
Closing due to inactivity; most of this just went in as part of #156

PR #149, opened 2022-01-02
add test first for matching with price deduction

cookie223:


jbms:
Thanks for adding this test --- but it appears that it does not pass.

cookie223:
These transactions look the same but somehow matching.py won't match them as the same. Probaly more changes than just verifying missing unit is needed. Will give it another try in the future.

PR #158, opened 2022-02-21
Remove bumpversion cfg.

carljm:
@jbms Is this obsolete? It looks like its purpose is to update version in setup.py, but that seems unnecessary now that it is tag-based.


jbms:
Yes it is obsolete.  Thanks for fixing that.

PR #159, opened 2022-02-22
Fix OFX transfer in with no unit price.

carljm:
These transfers can occur with no unit price; adjust the code to be able to handle that, since we don't actually expect to have a value for it in the transfer-in case anyway.


jbms:
Thanks

Issue #40, opened 2019-12-26
Price Format Support


ikwtif:
I was loading paypal data when I noticed price data doesn't take into account another format: dot before decimals (default) VS comma before decimals (drops the comma).
`XX.XX = XX.XX`
`XX,XX = XXXX` 
example: 
loaded data: `"price": "55,33\u00a0EUR"`
output: `5533 EUR`

 
This happens in `/beancount_import/amount_parsing.py` 
inside `def parse_amount(x)` 
at line 42 `number = D(m.group(2))` 
```
def parse_amount(x):
    """Parses a number and currency."""
    if not x:
        return None
    sign, amount_str = parse_possible_negative(x)
    m = re.fullmatch(r'([\$€£])?((?:[0-9](?:,?[0-9])*|(?=\.))(?:\.[0-9]+)?)(?:\s+([A-Z]{3}))?', amount_str)
    if m is None:
        raise ValueError('Failed to parse amount from %r' % amount_str)
    if m.group(1):
        currency = {'$': 'USD', '€': 'EUR', '£': 'GBP'}[m.group(1)]
    elif m.group(3):
        currency = m.group(3)
    else:
        raise ValueError('Failed to determine currency from %r' % amount_str)
    number = D(m.group(2))
    return Amount(number * sign, currency)
``` 

Is it possible to add a check for replacing `,` with `.` Not sure what the best method would be but I'm currently using this
```
    value = m.group(2)
    if ',' in value:
        value = value.replace(',','.')
    number = D(value)
```

jbms:
It would be good to support that, but it would have to somehow be a user-specified option, I think, since unfortunately the two styles are ambiguous.  The common style in the United States is to use commas as thousands separators:

"$1,999"  = 1999 USD
"$1,999.33"  =  1999.33 USD

TiraelSedai:
`XX.XX = XX.XX
XX,XX = XXXX`
This is a valid decimal conversion that is culture invariant and default in Python. Replacing `,` with `.` would only bring many issues.

In our country we have a comma instead of a dot as a decimal separator, so I know what I'm talking about.

What I would suggest is that you preprocess your Paypal data before feeding it into beancount-import.

mcbloch:
Why not pass the number into the Decimal class from `beancount.core.amount` before putting it into `Amount`? The Decimal class supports numbers with comma's. 

https://github.com/beancount/beancount/blob/master/beancount/core/number.py

moritzj29:
[...]

PR #161, opened 2022-03-11
Add workaround for new PayPal amounts format

rkok:
Adds a parsing fix for PayPal amounts observed to have a new format which looks like: '5.000,00\xa0EUR' (dot as thousands separator, comma as decimal separator, Unicode whitespace) which is not accepted by `parse_amount()`.

I don't know whether this format is specific to my locale or to all PayPal users.

I considered fixing this on the `finance-dl` side, but chose for this instead, because:
- It would mean unpacking and altering the JSON before writing it, which is currently not the case (seems elegant)
- There is no mechanism in place for fixing PayPal JSON files downloaded earlier AFAIK

Also decided against changing the regex in `parse_amount()` as it's already huge!

### DIFF
--- a/beancount_import/source/paypal.py
+++ b/beancount_import/source/paypal.py
@@ -402,9 +402,18 @@ class PaypalSource(LinkBasedSource, Source):
             tags=EMPTY_SET,
             postings=[])
         is_credit = data['isCredit']
-        counterparty_amount = parse_amount(data['amount']['grossAmount'])
-        funding_source_amount = parse_amount(data['amount']['netAmount'])
-        fee_amount = parse_amount(data['amount']['feeAmount'])
+
+        # Workaround for PayPal amounts observed to have a new format:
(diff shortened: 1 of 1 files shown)


moritzj29:
Hi, nice to see another `EUR` guy around here! 😄 If your are interested, check out #142 where I implemented a similar approach along the possibility to use Paypal data scraped from Paypal Germany. It should be easily extendible if you are interested.
Also for reference, let me mention #40 .

moritzj29:
Hi, nice to see another `EUR` guy around here! 😄 If your are interested, check out #142 where I implemented a similar approach along the possibility to use Paypal data scraped from Paypal Germany. It should be easily extendible if you are interested.
Also for reference, let me mention #40 .

Zburatorul:
@rkok, thanks for the contribution. It turns out moritzj29 was already working on something similar (and bigger in scope), so I'm closing this PR in favor of his.

Issue #143, opened 2021-12-11
[Paypal] SEND_MONEY_RECEIVED with fundingSource set breaks import


moritzj29:
Hi everyone,

I'm facing some issues with `SEND_MONEY_RECEIVED` PayPal transactions. In my case, all the transactions have a `fundingSource` entry in the JSON, therefore the import breaks at this line.

https://github.com/jbms/beancount-import/blob/dca340694ac60dd8ee7529cc4740b09fe729dbc2/beancount_import/source/paypal.py#L547

Since the only function of the line is to make sure NO `fundingSource` is set, I wonder what the purpose is. Removing the line, my imports work fine and I don't see any obvious problem yet.

For reference, I use transaction data from Paypal DE (#142) so maybe Paypal is doing some things differently in Germany...

moritzj29:
OK, thanks for your feedback! so I will just remove the line in my PR...

Zburatorul:
@jbms, it's not clear whether removing that line would break anything. Do you recall a reason for having it there?

jbms:
PayPal transactions sometimes apply to the PayPal balance itself, and sometimes bypass the PayPal balance and operate on external accounts directly.

On the line in question we are checking that there is no fundingSource since we are creating a transaction against the PayPal balance itself, whereas the presence of a funding account would suggest that the transaction should be operating on that external account directly.

Can you confirm that the transaction in question is in fact operating on the PayPal balance directly, not on the external account?  What is the fundingSource indicating?

moritzj29:
my transactions of type `SEND_MONEY_RECEIVED` have a funding source specified, which is named `BALANCE`. Since the `fundingSource` is given, the `assert` fails.

```
{
  "fundingSource": {
    "fundingSourceList": [
      {
        "type": "BALANCE",
        "amount": "15,00\u00a0EUR",
        "currencyCode": "EUR"
      }
    ]
  },
```

Private Paypal transactions working on bank accounts directly appear to have the type `SEND_MONEY_SENT`

jbms:
Thanks --- looks like removing the assert is indeed correct.

PR #163, opened 2022-03-22
[Paypal] remove check on no funding source for SEND_MONEY_RECEIVED

moritzj29:
as discussed in #143.

I think it's better to keep this separated from all the other changes in #142 ...


moritzj29:
merged with #142

PR #164, opened 2022-03-26
[Venmo] Handle new transaction type - "Merchant Transaction".

Zburatorul:
I've seen `Merchant Transaction` appear among my transactions. Venmo seems to apply this to some in-app payments (e.g. Uber Eats).


jbms:
Thanks!

PR #167, opened 2022-05-15
Generic Importer: Respect tolerance during balancing

msiedlarek:
Account for rounding errors by adding Expenses:FIXME posting only when the amount is smaller then the deduced/configured tolerance.

Without this change the added test case would have received a posting `Expenses:FIXME 0.000100 USD`.


addisonklinke:
Running into this as well (particularly for stock transactions). Is the merge just blocking for tests?

PR #173, opened 2022-08-22
ofx: vendor parse_ofx_time implementation

jktomer:
The ofx source makes use of a simple function to parse the OFX time format
from an OFX importer that used to be included in Beancount. Unfortunately that
importer has been demoted to a separate repository `beangulp`, and to the
examples directory in that repository to boot. This change therefore copies
the (trivial) implementation of `parse_ofx_time` from the beangulp examples
into ofx.py.


jbms:
Thanks

PR #174, opened 2022-09-02
Improve matching by using the sliding slice approach

xentac:
Here are the improvements that I've come up with so far. It can still be slow, but it's definitely an improvement over some of the algorithms before.


beancount_import/matching.py
```diff
@@ -240,6 +254,20 @@ def get_fuzzy_date_range(self, orig_date: datetime.date):
                                 self.fuzzy_match_days + 1):
             yield orig_date + datetime.timedelta(days=day_offset)
 
+    def fuzz_date_currency_key(self, key: Tuple):
```
jbms, line 257: Type annotation should be `DateCurrencyKey` rather than `Tuple`, I think

jbms: Also please add return annotation, e.g. `Iterator[DateCurrencyKey]`

beancount_import/matching.py
```diff
@@ -240,6 +254,20 @@ def get_fuzzy_date_range(self, orig_date: datetime.date):
...
+        for d in self.get_fuzzy_date_range(key[0]):
+            yield d, key[1]
+
+    def _date_currency_key(self, entry: Transaction, mp: MatchablePosting):
```
jbms, line 261: please add return type annotation

beancount_import/matching.py
```diff
@@ -255,6 +283,81 @@ def add_posting(self, entry: Transaction, mp: MatchablePosting):
         group = self._postings.setdefault(_date_key(entry, mp), {})
         group[source_posting_ids] = (entry, mp)
 
+        weight = get_posting_weight(mp.posting)
+        if weight is None:
+            return
+
+        for dc in self.fuzz_date_currency_key(self._date_currency_key(entry, mp)):
+            bisect.insort(
```
jbms, line 291: It would be better to lazily-sort these lists, i.e. add to the end when inserting, and record that the list is not sorted.  Then when querying, sort the list if it isn't already sorted.

When using `insort` as you have here, adding `n` postings with the same currency and date is `O(n**2)`.

beancount_import/matching.py
```diff
@@ -772,6 +880,11 @@ def add_subset(account, currency, subset, check_zero=True):
                 2, min(len(posting_list) + 1, max_subset_size + 1)):
             for subset in itertools.combinations(posting_list, subset_size):
                 add_subset(account, currency, subset)
+
```
jbms, line 883: We already loop over subset_size just above here, so it would better to re-order the looping above so that we can avoid spending the time generating subsets that we will just throw away anyway.

beancount_import/matching.py
```diff
@@ -255,6 +283,81 @@ def add_posting(self, entry: Transaction, mp: MatchablePosting):
...
+                                    # Stop searching if we match a keyed posting
+                                    newsp = SearchPosting(
+                                        number=None, key=k, entry=t, mp=mp)
+                                    if cmp_callable(p, newsp):
+                                        yielded = True
+                                        yield newsp
+                        if yielded:
+                            continue
+
+                upper = weight.number + self.fuzzy_match_amount
```
jbms, line 342: I don't think there is necessarily much gain in trying to do multiple posting queries as a batch here.  I think if you just did a binary search it would be plenty fast and would simplify the code --- basically the same structure as before could be used.

beancount_import/matching.py
```diff
@@ -255,6 +283,81 @@ def add_posting(self, entry: Transaction, mp: MatchablePosting):
...
+        # Prepare postings for searching
+        postings_date_currency = collections.defaultdict(list)
+        
+        for mp in mps:
+            weight = get_posting_weight(mp.posting)
+            if weight is None:
+                continue
+            if negate:
+                weight = -weight
+            bisect.insort(
```
jbms, line 309: Definitely should avoid using `bisect.insort` here (instead could sort all of the lists at the end), but if you change to just looking up each posting match individually, there is no need for this data structure at all.

Possibly with this issue fixed you could increase the limit on number of matching postings to allow, since as it is currently, it is O(n**2), where n is the number of matchable postings with a given (date, currency).

jbms:
Thanks very much for working on this!

beancount_import/matching.py
xentac: I ran both implementations side by side and was surprised to find that you're right :)

I'm working on removing the new code I created and adopting the bisect implementation.

xentac: After fiddling with the code some more and clearing some bugs, it turns out I was wrong.

My bug was I was always finding 0 results during matching. Turns out that's faster than actually matching things.

Current implementation wise, the bisecting takes 2.1 seconds (time spent in _get_matches), the sliding iterators take 0.93 seconds (time spent in search_postings).

xentac:
Please take a look at what I have and tell me if there are other things you're looking for.

This does make my matching faster, but it ends up finding a lot of spurious matches. Part of the problem is the large transactions have a lot of subsets that match a lot of transactions in my personal file.

It is complicated even more because I want it to find the two negative postings (the main payment and the tip, in my Amazon grocery orders). The others I rarely care about. Apart from just giving up on the super large transactions and only searching for only the negative transactions, I'm not sure the right way to approach this.

beancount_import/matching.py
```diff
@@ -220,8 +227,17 @@ def _entry_and_posting_ids_key(entry: Transaction, mp: MatchablePosting):
 # Keyed by the sequence of source_posting ids.
 DatabaseValues = Dict[SourcePostingIds, Tuple[Transaction, MatchablePosting]]
 
+SearchPosting = NamedTuple('SearchPosting', [
+    ('number', Decimal),
+    ('key', Tuple),
```
jbms, line 232: Please specify Tuple element types

beancount_import/matching.py
```diff
@@ -240,6 +258,28 @@ def get_fuzzy_date_range(self, orig_date: datetime.date):
                                 self.fuzzy_match_days + 1):
             yield orig_date + datetime.timedelta(days=day_offset)
 
+    def fuzz_date_currency_key(self, key: DateCurrencyKey) -> Iterable[DateCurrencyKey]:
+        for d in self.get_fuzzy_date_range(key[0]):
+            yield d, key[1]
+
+    def _date_currency_key(self, entry: Transaction, mp: MatchablePosting) -> DateCurrencyKey:
+        pw = get_posting_weight(mp.posting)
+        currency = ""
```
jbms, line 267: Is it possible to have a missing currency for a matchable posting?

beancount_import/matching.py
```diff
@@ -231,6 +247,8 @@ def __init__(self, fuzzy_match_days: int,
         self.fuzzy_match_amount = fuzzy_match_amount
         self.is_cleared = is_cleared
         self._postings = {}  # type: Dict[DatabaseDateKey, DatabaseValues]
```
jbms, line 249: I think we can eliminate `self._postings` since it is redundant with `self._date_currency`

beancount_import/matching.py
```diff
@@ -240,6 +258,28 @@ def get_fuzzy_date_range(self, orig_date: datetime.date):
...
+    def _date_currency_key(self, entry: Transaction, mp: MatchablePosting) -> DateCurrencyKey:
+        pw = get_posting_weight(mp.posting)
+        currency = ""
+        if pw:
+            currency = pw.currency
+        return (_date_key(entry, mp), currency)
+
+    def _search_posting(self, entry: Transaction, mp: MatchablePosting):
+        pw = get_posting_weight(mp.posting)
+        number = None
```
jbms, line 274: number should also always be present for a matchable posting

beancount_import/matching.py
```diff
@@ -255,6 +295,86 @@ def add_posting(self, entry: Transaction, mp: MatchablePosting):
         group = self._postings.setdefault(_date_key(entry, mp), {})
         group[source_posting_ids] = (entry, mp)
 
+        weight = get_posting_weight(mp.posting)
+        if weight is None:
```
jbms, line 299: I think `weight` should never be None for a MatchablePosting

beancount_import/matching.py
```diff
@@ -255,6 +295,86 @@ def add_posting(self, entry: Transaction, mp: MatchablePosting):
...
+            return
+
+        for dc in self.fuzz_date_currency_key(self._date_currency_key(entry, mp)):
+            self._date_currency[dc].append(self._search_posting(entry, mp))
+            self._date_currency_dirty[dc] = True
+
+    def get_date_currency_postings(self, key: DateCurrencyKey) -> List[SearchPosting]:
+        if self._date_currency_dirty[key]:
+            self._date_currency[key].sort()
```
jbms, line 308: could save one dict lookup by setting:

results = self._date_currency[key] before the if statement

beancount_import/matching.py
xentac: Done.

beancount_import/matching.py
xentac: I have addressed this in my current code.

beancount_import/matching.py
xentac: It's the output of _entry_and_posting_ids_key... I think that makes it a SourcePostingIds.

beancount_import/matching.py
xentac: I have no idea. I don't understand the loader code well enough to know.

beancount_import/matching.py
xentac: I agree. I had left it in case I missed something :) I'll follow up with a change to remove it.

beancount_import/matching.py
xentac: This one is actually tricky. I use self._date_currency_dirty to tell me if I need to sort self._date_currency. They are different dicts.

beancount_import/matching.py
xentac: See https://github.com/jbms/beancount-import/pull/174#discussion_r1089647813

beancount_import/matching.py
xentac: See https://github.com/jbms/beancount-import/pull/174#discussion_r1089647813

xentac:
I have responded to your comments and added a few more changes. Now one of my invoices that took forever (with some of the improvements it took 7000 seconds) is now down to 240 seconds. There's still room for improvement in places, but it's getting harder to optimize some of these stateful methods.

Please take another look.

xentac:
I believe I've addressed all the comments and lint errors. Please take a look.

I'm sure there are more improvements that can be made, but they're definitely getting harder to find.

Large invoices still take hundreds of seconds, which is an improvement to the thousands/infinite seconds it took previously.

xentac:
I think the next area I will look is probably compute_balanced_match_group and compute_single_sign_match_groups.

Another place that I feel like could be improved is in is_metadata_mergeable, but I don't quite know how to resolve the question of mergability without using mutations and trying.

beancount_import/matching.py
jbms: Ah, missed that, sorry.

jbms:
Thanks, this is a great improvement, and we can iterate on it further in the future if needed.

Zburatorul:
@xentac, thank you for this contribution. I look forward to not waiting forever on my multi-posting transactions.

beancount_import/matching.py
```diff
@@ -318,24 +435,30 @@ def _get_matches(
...
-                        continue
+        dc = (date, amount.currency)
+        upper = amount.number + self.fuzzy_match_amount
+        lower = amount.number - self.fuzzy_match_amount
+
+        cur_matches = self.get_date_currency_postings(dc)
+        if cur_matches is not None:
+            lower_bound = bisect.bisect_left(cur_matches, (lower, tuple(), None, None))
+            upper_bound = bisect.bisect_right(cur_matches, (upper, (sys.maxsize,), None, None), lo=lower_bound)
+            for sp in cur_matches[lower_bound-1 if lower_bound > 0 else 0:upper_bound]:
```
ankurdave, line 446: @xentac It seems like this `lower_bound-1` adjustment will bring in an extra element that falls outside the bounds, causing https://github.com/jbms/beancount-import/pull/206#issuecomment-1789816679. Do you remember why it was necessary?

xentac:
It's entirely possible that it was a bug. I may have been confused about
the return value of bisect_left and it being the insertion point.

PR #176, opened 2022-09-27
Add support for options transactions

naskooskov:
This patch started as fixing issue 175, though I realized it needs a bit more support in order to handle options correctly. Not sure if this is idiomatic python or whether it fits well with the project style, so happy to receive comments.


jbms:
Thanks.  Can you rebase to confirm that tests pass?

naskooskov:
Rebased and tests pass locally.

Zburatorul:
`FAILED beancount_import/source/ofx_test.py::test_source[test_fidelity_ira-fidelity-ira.ofx]`

naskooskov:
Added a couple of files I had missed in the initial commit.

naskooskov:
Another missed file ...

beancount_import/source/ofx.py
```diff
@@ -967,10 +979,28 @@ def get_subaccount_cash(inv401ksource: Optional[str] = None) -> str:
 
                 cost_spec = None
                 price = None
-                is_sale = False
-                if raw.trantype in SELL_TYPES or (raw.trantype == 'TRANSFER' and
+                is_closing_txn = False
+                if raw.trantype in OPT_TYPES:
+                    assert(raw.optsharesperctrct is not None)
+
+                    if raw.opttradetype.find('TOCLOSE') == -1:
```
Zburatorul, line 986: Have you considered adding support for the `CLOSUREOPT` transaction type as well?
Not saying PR needs it, but it might be an easy addition.

testdata/source/ofx/test_fidelity_ira/journal.beancount
```diff
@@ -0,0 +1,139 @@
...
+1900-01-01 open Assets:Investment:Fidelity:IWM220826C207
+1900-01-01 open Assets:Investment:Fidelity:MSFT220826C310
+1900-01-01 open Assets:Investment:Fidelity:WMT220617P118
+1900-01-01 open Expenses:Investment:Fidelity:Fees
+1900-01-01 open Expenses:Investment:Fidelity:Commissions
+1900-01-01 open Expenses:FIXME
+1900-01-01 open Income:Fidelity:Capital-Gains
+1900-01-01 open Income:Fidelity:Capital-Gains:DAL220520C42
+
+2022-05-13 * "SELLOPT - DAL220520C42"
```
Zburatorul, line 28: Can we add an extra keyword in the description to indicate whether it's sell to open or sell to close?
I know it's in the metadata, but it'd make it easier to understand activity at a glance. Plus the description field is pretty short at the moment, so no downside.

testdata/source/ofx/test_fidelity_ira/journal.beancount
```diff
@@ -0,0 +1,139 @@
...
+    cleared: TRUE
+
+2022-06-17 * "TRANSFER - WMT220617P118"
+  Assets:Investment:Fidelity:WMT220617P118  1 WMT220617P118 {}
+    date: 2022-06-17
+    ofx_fitid: "31337997217301320220617"
+    ofx_memo: "EXPIRED PUT (WMT)"
+    ofx_type: "TRANSFER"
+    cleared: TRUE
+  Expenses:FIXME
```
Zburatorul, line 86: For the buy to close example above I notice you generated an income posting against the right account automatically. Is there a reason not to do the same here?

testdata/source/ofx/test_fidelity_ira/journal.beancount
```diff
@@ -0,0 +1,139 @@
...
+
+2022-08-18 * "SELLOPT - MSFT220826C310"
+  Expenses:Investment:Fidelity:Fees          0.0400 USD
+  Expenses:Investment:Fidelity:Commissions   0.5000 USD
+  Assets:Investment:Fidelity:MSFT220826C310      -1 MSFT220826C310 {10.0} @ 0.1 USD
+    date: 2022-08-18
+    ofx_fitid: "31337997223101120220818"
+    ofx_memo: "YOU SOLD OPENING TRA"
```
Zburatorul, line 103: I only see two kind of examples: sell to open, buy to close.
We want the other two as well (unless they are already somewhere?): sell to close, buy to open.

Zburatorul (commented):
Looking good. Some clarifying questions from me.

beancount_import/source/ofx.py
```diff
@@ -967,10 +979,28 @@ def get_subaccount_cash(inv401ksource: Optional[str] = None) -> str:
...
+                        price_per = unitprice * raw.optsharesperctrct
```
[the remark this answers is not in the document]
naskooskov: Here is a bit of the logic behind this code. Each option contract corresponds to some number of shares, usually 100. optsharesperctrct contains that value, so to get the price one actually pays, we need to multiply. Let's look at an example to make sure we are on the same page.
Let's say I trade 1 contract for $0.53. Since the contract corresponds to 100 shares, the actual cost for the transaction is 0.53 * 100 or $53. This is why I name this price_per ... since price is already taken and this is the real money we have transacted for the contract. I'm happy to name it price_per_contract, though in our example this would usually be $0.53. Open to suggesstions since naming is a hard problem :).

beancount_import/source/ofx.py
naskooskov: I don't know that I have a real example from real world ofx file, so I've only added code for transactions I've seen.

beancount_import/source/ofx.py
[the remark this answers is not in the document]
Zburatorul: I would name this price_per_batch.

Zburatorul:
Thank you for the contribution @naskooskov

PR #177, opened 2022-10-15
[Google] Allow exactly one dot at the end of line.

delphij:
In 2022, accidential insurance is represented as "Accidential Ins." on statements, this will allow these to be used.


delphij:
> Thanks. Can you rebase to confirm that tests pass?

Sure.  I've also added a new test case for it.

PR #180, opened 2022-12-18
fix: replace `get_feature_names` with `get_feature_names_out` for scikit-learn classifier

JonBoyleCoding:
I'm not sure exactly how this broke for me as this worked perfectly fine for me a month ago - perhaps a system update updated `scikit-learn`. 

It seems `scikit-learn 1.2.0`  uses `get_feature_names_out` for the classifier instead of:

 https://github.com/jbms/beancount-import/blob/afd3093ff3e1cf35663f0ba170e2f94451fc90b0/beancount_import/reconcile.py#L139

At least the commit I made was necessary for it to work on my machine. I made this change for myself and all seems to work - presumably it's needed here so I thought to make the PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,9 +11,9 @@ jobs:
           - 'windows-latest'
           - 'macos-latest'
         python-version:
-          - '3.7'
           - '3.8'
           - '3.9'
+          - '3.10'
         node-version:
           - '15.x'
     runs-on: ${{ matrix.os }}
--- a/beancount_import/reconcile.py
+++ b/beancount_import/reconcile.py
@@ -136,7 +136,7 @@ def get_prediction_explanation(classifier, features: Dict[str, bool]):
 
     converted_features = classifier._vectorizer.transform([features])
     class_names = classifier._encoder.classes_
-    feature_names = classifier._vectorizer.get_feature_names()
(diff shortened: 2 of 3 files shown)


m-d-brown:
I just encountered this as well after installing the newest version of scikit-learn. The fix works for me as well.

JonBoyleCoding:
I've added a commit with a version tag now - hopefully this is ok?

JonBoyleCoding:
I noticed the build failed - apparently scikit-learn 1.2.0 only supports `python>=3.8`, and the build that failed was the 3.7.

https://pypi.org/project/scikit-learn/

JonBoyleCoding:
I could remove `3.7` from the workflow, and add `3.10`? I'm using `3.10` and it seems to work fine.

Zburatorul:
Can you rebase so we can run the build again. Should work now.

JonBoyleCoding:
Sure - I will look at this sometime today or this evening.

JonBoyleCoding:
Apologies it took so long - a lot going on. FIngers crossed this will work!

Zburatorul:
I'm okay with removing python3.7 as it's nearly EOL: https://devguide.python.org/versions/
Hopefully this doesn't cause others too much trouble.

amundhov:
Any chance we could make a tagged release to pypi including this PR?

PR #182, opened 2023-01-07
fix mypy errors of implicit optional arguments

moritzj29:
change of default mypy option according to PEP484

should fix CI pipeline


Zburatorul:
Finally it's unblocked! Thank you!

PR #183, opened 2023-01-07
make parse_amount work with currency symbols after amount

moritzj29:
also allow symbols after the amount (currently only 3 letter currency possible)
allows e.g. `2.99 €` instead of `2.99 EUR`

required since Paypal changed their website and therefore also the JSON fetched via `finance-dl` changed to this notation.


Zburatorul:
pipeline failing ever after latest merge

Zburatorul:
Ah yes, a rebase was needed. Thank you.

PR #186, opened 2023-02-06
Allow specifying custom currency in Mint importer

jasonhuhx:
This PR enables custom currency for Mint importer.

Users can now specify any currency in data source:

```python
    data_sources = [
        dict(
            module='beancount_import.source.mint',
            filename=os.path.join(data_dir, 'mint/mint.csv'),
            currency='CAD',
        )
    ]
```


beancount_import/source/__init__.py
```diff
@@ -255,7 +255,6 @@ class Source:
     """
 
     def __init__(self, log_status: Callable[[str], None], **kwargs) -> None:
-        super().__init__(**kwargs)  # type: ignore
```
Zburatorul, line 258: Why was this removed?

Zburatorul (commented):
Can you also a test by creating another test method like this one:
https://github.com/jbms/beancount-import/blob/59615a5857303fdfa310d930bbe96f12ed29e411/beancount_import/reconcile_test.py#L231

You'll also have to create a CAD version of that test data.

beancount_import/source/__init__.py
jasonhuhx: When we pass in an additional parameter (for example `currency` in this case) to the data source initialization dictionary, thie line reports the following error:

```
  File "/Users/jason/.pyenv/versions/3.8.11/lib/python3.8/site-packages/beancount_import/source/__init__.py", line 258, in __init__
    super().__init__(**kwargs)
TypeError: object.__init__() takes exactly one argument (the instance to initialize)
> /Users/jason/.pyenv/versions/3.8.11/lib/python3.8/site-packages/beancount_import/source/__init__.py(258)__init__()
-> super().__init__(**kwargs)
```

This is not specific to Mint importer. Adding any additional parameter to any the the importer will cause this error. Since the class `Source` doesn't have any parent class, it should be fine to remove this line.

Zburatorul: I don't like dropping the base class init call completely, but I can live with `super().__init__()`.

testdata/reconcile/test_currency/0/candidates.diff
```diff
@@ -0,0 +1,20 @@
+<journal-dir>/journal.beancount
```
Zburatorul, line 1: Upon closer inspection I realized that `mint_test.py` might be a more appropriate place to put the currency test. Could you move your test code into it, as well as the test artifacts under `testdata/source/mint/test_currency`.

Sorry I misguided you initially.

jasonhuhx: moved tests to the appropriate folder

beancount_import/source/__init__.py
jasonhuhx: Now keeping `super().__init__()`

Zburatorul:
Merged. Thanks!

PR #187, opened 2023-02-06
Add ofx support for treasury T-Bill

zhou13:
* Add ofx support for treasury T-Bill
* Workaround the issues that Fidelity OFX shows the units of treasuries 100x larger.


beancount_import/source/ofx.py
```diff
@@ -785,6 +786,10 @@ def get_security(unique_id: str) -> Optional[str]:
                 return None
             sec = securities_map[unique_id]
             ticker = sec.ticker
+            # Treasury bill and bond start with 912
+            if ticker.startswith("912"):
+                # Append "T" to make it a valid ticker
```
Zburatorul, line 791: Was this meant to be prepend?

beancount_import/source/ofx.py
```diff
@@ -785,6 +786,10 @@ def get_security(unique_id: str) -> Optional[str]:
                 return None
             sec = securities_map[unique_id]
             ticker = sec.ticker
+            # Treasury bill and bond start with 912
+            if ticker.startswith("912"):
```
Zburatorul, line 790: Is the ticker of a bond always the same as its CUSIP?

beancount_import/source/ofx.py
```diff
@@ -964,6 +969,9 @@ def get_subaccount_cash(inv401ksource: Optional[str] = None) -> str:
                     units = round(units, quantity_round_digits)
                 if raw.unitprice is not None:
                     unitprice = normalize_fraction(raw.unitprice)
+                # For fidelity ofx, the unit of bond has been multiplied by 100
+                if self.org == "fidelity.com" and raw.uniqueid.startswith("912") and units % 100 == 0:
```
Zburatorul, line 973: I wonder if we make this more general and future-proof by detecting that a rescaling is required.
I ran into a similar/same issue a while ago, and my solution was to keep the units but rescale the price.
I detected the problem by noticing that `market value != units * price`.
Can you see if adopting my solution solves the problem?

```@@ -714,12 +806,23 @@ class ParsedOfxStatement(object):
                 time_str = find_child(invpos, 'dtpriceasof')
                 t = parse_ofx_time(time_str)
                 date = t.date()
+                units=find_child(invpos, 'units', D)
+                unitprice=find_child(invpos, 'unitprice', D)
+                mktval=find_child(invpos, 'mktval', D)
+                if mktval and mktval > 0:
+                    ratio = units*unitprice/mktval
+                    # these thresholds are arbitrary and could be tightened
+                    if (ratio < 0.2) or (ratio > 5):
+                        id_type = find_child(invpos, 'uniqueidtype')
+                        unique_id = find_child(invpos, 'uniqueid')
+                        unitprice = mktval / units if units > 0 else None
+                        print(
+                            f"[{id_type} {unique_id}]: Mismatch between UNITS * UNITPRICE = {units*unitprice:.2f} and  MKTVAL = {mktval:.2f}. Inferring price: {unitprice:.3f}")
                 raw_balance_entries.append(
                     RawBalanceEntry(
                         date=date,
                         uniqueid=find_child(invpos, 'uniqueid'),
                         units=find_child(invpos, 'units', D),
-                        unitprice=find_child(invpos, 'unitprice', D),
+                        unitprice=unitprice,
                         inv401ksource=find_child(invpos, 'inv401ksource'),
                         filename=filename))```

**Review by Zburatorul — COMMENTED** [review]
@zhou13, thank you for identifying a real gap in `ofx.py` and closing it.

File: beancount_import/source/ofx.py (hunk 2 above)
Comment (zhou13) [reply beancount_import/source/ofx.py:790]: It is at least for Fidelity's Ofx. I don't have other OFX so I am not sure about it.

File: beancount_import/source/ofx.py (hunk 1 above)
Comment (zhou13) [reply beancount_import/source/ofx.py:791]: Yes, fixed.

File: beancount_import/source/ofx.py (hunk 3 above)
Comment (zhou13) [reply beancount_import/source/ofx.py]: Thanks, this looks good!

**Comment by zhou13** [comment]
Pytest runs fine locally, not sure what happenes.

**Comment by Zburatorul** [comment]
> Pytest runs fine locally, not sure what happens. Also the balance change make a lot of changes to the existing tests.

That's exactly what happens. We need to figure out if the changes are correct or not.

Here is the [OFX specification](https://financialdataexchange.org/common/Uploaded%20files/OFX%20files/OFX%20Banking%20Specification%20v2.3.pdf) that describes the meaning of each field in detail.

For `UNITS` it says:
```
For stocks, MFs, other, number of shares held.
Bonds = face value.
Options = number of contracts
```
Meanwhile for `UNITPRICE` it says`Bonds = percentage of par`. Meaning a number from 0 to 100, not a dollar amount.
We could just ignore all this and record `UNITPRICE` as the price, but it would produce incorrect (about 100X inflated) asset values. This also explains why Fidelity has that 100X factor from your original commit.

There are two ways out: either we modify the amount, or the price.
I'm leaning towards transforming the price to match beancount't expectation of what a price is (market value/quantity of security) rather than modifying the quantity/units.
There's a subcase where we detect that it's a bond, and simply divide the price by 100X. But my suggested code takes care of that AND of situations where the multiplier is not 100.

There are a number of ofx test files where `market value != price * units`. That is why the tests are failing.
For example `testdata/source/ofx/vanguard_roth_ira.ofx` has:
```
  <UNITS>46.872
  <UNITPRICE>84.20
  <MKTVAL>72.85
```
But it's not difficult to fix them, and it would result in a PR that is more general and more robust, in my opinion.
The work consists of inspecting each failed test and seeing that 1) no balance entries are modified 2) the new price indeed produces the correct market value.

We're really close here. Are you willing to get it over the finish line?

**Comment by zhou13** [comment]
Do you know how to reproduce the CI error locally? Currently running
```
pytest -s beancount_import/source/ofx_test.py
```
give me no error.

**Comment by Zburatorul** [comment]
Maybe you have a local file that that you forgot to check in?

**Comment by Zburatorul** [comment]
Here's a sample error:
>E             - 2018-07-03 price TYCDT                  1.554232804232804232804232804 USD
E             + 2018-07-03 price TYCDT                               84.20 USD

It indicates that you changed the price, but did not modify the expected output file. Please double-check that the new price is actually correct.

**Comment by zhou13** [comment]
Here is what I did:
```
git clone https://github.com/zhou13/beancount-import bc2
cd bc2
git checkout fidelity-buydebt
pip install git+https://github.com/zhou13/beancount-import@fidelity-buydebt
pytest -s beancount_import/source/ofx_test.py
```
It passes all the test. The change `E - 2018-07-03 price TYCDT 1.554232804232804232804232804 USD` is from executing `check_source_example(..., write=True)` on my local machine. I don't understand why the result is different from the CI.

**Comment by Zburatorul** [comment]
Yeah it is strange. I don't understand why the file in the PR is different from the one in the test.
Try `git push --force` once again, and let's re-run the test.

If nothing changes, then I suggest you revert the `testdata/source/ofx/test_vanguard_roth_ira_matching/import_results.beancount` back to master, commit & force push that. We might get a new clue from that test passing/failing in CI.

**Comment by zhou13** [comment]
@Zburatorul I think this PR is ready to go. Previously I missed a test, which is why it fails. I checked `investment_buy_sell_income.ofx` and the new result is more correct as `MKTVAL` is a more reliable entry.

File: beancount_import/source/ofx.py [hunk 4]
```diff
@@ -617,13 +618,17 @@ def get_securities(soup: bs4.BeautifulSoup) -> List[SecurityInfo]:
...
+    ['BUYMF', 'SELLMF', 'SELLSTOCK', 'BUYSTOCK', 'REINVEST', 'BUYDEBT',
+     'SELLDEBT', 'SELLOTHER'])
+SELL_TYPES = set(['SELLMF', 'SELLSTOCK', 'SELLDEBT', 'SELLOTHER'])
 
 RELATED_ACCOUNT_KEYS = ['aftertax_account', 'pretax_account', 'match_account']
 
 # Tolerance allowed in transaction balancing.  In units of base currency used, e.g. USD.
 TOLERANCE = 0.05
+UNITPRICE_ERROR_LOWER_BOUND = 0.05
+UNITPRICE_ERROR_UPPER_BOUND = 20.0
```
Zburatorul, line 630: I think I had 5 and 1/5 originally which I thought was already far too conservative. What's the logic for making these choices even more conservative? Do you have a example where the 5 and 1/5 bounds lead to false positives?

beancount_import/source/ofx.py
```diff
@@ -676,6 +681,17 @@ def __init__(self, seen_fitids, filename, securities_map, org, stmtrs):
...
                     total = find_child(tran, 'total', D)
+                units = find_child(tran, 'units', D)
+                unitprice = find_child(tran, 'unitprice', D)
+                if total and unitprice:
+                    error_ratio = abs(units * unitprice / total)
+                    if error_ratio > UNITPRICE_ERROR_UPPER_BOUND or error_ratio < UNITPRICE_ERROR_LOWER_BOUND:
+                        id_type = find_child(tran, 'uniqueidtype')
+                        unique_id = find_child(tran, 'uniqueid')
+                        units_x_unitprice = units*unitprice
+                        unitprice = abs(total / units)
```
Zburatorul, line 692: Let's handle edge case where `units = 0` please.

testdata/source/ofx/fidelity.ofx
```diff
@@ -1,11 +1 @@
-OFXHEADER:100
-DATA:OFXSGML
-VERSION:102
-SECURITY:NONE
-ENCODING:USASCII
-CHARSET:1252
-COMPRESSION:NONE
-OLDFILEUID:NONE
```
Zburatorul, line 8: Why was this removed?

testdata/source/ofx/test_vanguard_roth_ira_matching/training_examples.json
```diff
@@ -1,13 +1 @@
...
-  [
-    {
-      "amount": "31.704 TYCDT",
-      "date": "2018-06-21",
-      "key_value_pairs": {
-        "ofx_type": "REINVEST"
-      },
-      "source_account": "Assets:Retirement:Vanguard:Roth-IRA:TYCDT"
-    },
-    "Income:Vanguard:Dividends:TYCDT"
```
Zburatorul, line 11: Why was this removed? I don't understand.

testdata/reconcile/test_ofx_matching/0/candidates.diff
```diff
@@ -0,0 +1,3 @@
+<journal-dir>/journal.beancount
++
++2018-07-03 price TYCDT                  1.554232804232804232804232804 USD
```
Zburatorul, line 3: In another PR someone should think carefully about how to truncate numbers like this correctly. They are an eye sore.

Zburatorul (commented):
A few questions

beancount_import/source/ofx.py
zhou13: I don't have an example. Right now I think the only case to trigger this logic is for bond, where the ratio is 100x. So a 20x ratio is enough. I don't object to making it more aggressive.

testdata/source/ofx/test_vanguard_roth_ira_matching/training_examples.json
zhou13: Restored.

beancount_import/source/ofx.py
zhou13: Done

testdata/source/ofx/fidelity.ofx
zhou13: Restored.

testdata/source/ofx/test_vanguard_roth_ira_matching/training_examples.json
zhou13: I added `div_income` to `AUX_ACCOUNT_KEYS`, which causes these changes. This is because for treasury bills, the difference between the price of buying and seller should be counted as dividends.

I actually don't understand the meaning of `AUX_ACCOUNT_KEYS` and why it is needed to ensure the account isn't used for training though.

beancount_import/source/ofx.py
Zburatorul: Yeah let's make this 5 and 1/5 then.

testdata/source/ofx/test_vanguard_roth_ira_matching/training_examples.json
Zburatorul: >is because for treasury bills, the difference between the price of buying and seller should be counted as dividends.
That doesn't sound right to me. What's the logic?

The price can change even if no coupon payments have been made. Regular capital gain/loss.

Zburatorul: re: `AUX_ACCOUNT_KEYS` it looks like the logic is that since the user has manually provided auxiliary account keys, they must know what they're doing, and b-i shouldn't try to learn or infer.

I do agree agree `div_income` makes sense as part of `AUX_ACCOUNT_KEYS`. @jbms, is there anything we're missing here?

zhou13: I think the confusion is because I forgot to include a test case. It has been added and it should be clear.

Here is the base case: Treasury T-Bill (in contrast to Treasury bond.  T-Bill doesn't pay any coupon) sells at a discount: For example, you buy a bill at $98.00 (`buydebt`) and when it matures, the Treasury pays you $100 (`sellother` in ofx). The $2 difference goes to interests. Previously, I made a mistake in thinking it was dividends. I have fixed it in the PR and as a result, `training_examples.json` has been restored.

Now, you can sell a T-Bill before it matures. According to the reference, it seems that you will have additional capital gain/loss in addition to the interests if the selling price differs from the mature price. This PR does not support this case because I haven't never done this (no ofx for testing) and I am not sure if my understanding is 100% correct.

Reference: https://public.com/learn/how-are-treasury-bills-taxed#are-t-bills-taxed

beancount_import/source/ofx.py
zhou13: I can verify this is also true for vanguard treasury.

zhou13:
@Zburatorul I have been using this branch for a while for Fidelity Treasury bill and it works well for me. I also add tolerance to workaround buggy rounding of the `unit` entry in Vanguard's bond ofx so it can generate reasonable balance. Let me know if you have any additional concerns.

Zburatorul:
Thank you for the contribution @zhou13!

beancount_import/source/ofx.py
```diff
@@ -797,6 +833,10 @@ def get_security(unique_id: str) -> Optional[str]:
                 return None
             sec = securities_map[unique_id]
             ticker = sec.ticker
+            # Treasury bill and bond start with 912
+            if ticker.startswith("912"):
```
ankurdave, line 837: FYI, this will crash if the ticker is `None`, rather than printing a helpful error message as it used to.

PR #195, opened 2023-03-12
Fix crashes when db[x].number is None

zhou13:


Zburatorul:
@zhou13 have you tried the latest master commit? I think it fixes the same issue.
If you still get crashes, we can figure out why.

zhou13:
I test the master and it has already been fixed. Thank you!

PR #198, opened 2023-03-19
fix parse error with empty price, fixes #197

### LINKED ISSUE #197: Amazon import fails when Price is empty
Seen Error Message:
error:
Failed to parse invoice amazon_order_xxxxx.html: (<class 'AttributeError'>, AttributeError("'NoneType' object has no attribute 'number'"), <traceback object at 0x10311ca00>)

When Items are exchanged, no price is listed, resulting in a None in for price in amazon_invoicy.py

concoy:


zhou13:
It seems `Locale_de_DE.currency` is hard coded. Does this affect users from other country?

Issue #202, opened 2023-06-25
transactions matched that shouldn't be


falsifian:
I'm on `master`, commit b8a78069, and getting the following candidate:

```
+
+2023-01-01 * 
+  Assets:Chequing   20.00 CAD
+    date: 2023-01-01
+    ofx_fitid: "0"
+    ofx_name: "Transaction 0"
+    ofx_type: "STMTTRN"
+  Assets:Chequing  -10.00 CAD
+    date: 2023-01-01
+    ofx_fitid: "1"
+    ofx_name: "Transaction 1"
+    ofx_type: "STMTTRN"
```

Notice the `20.00 CAD` is quite different from `-10.00 CAD`. I haven't changed the default `fuzzy_match_amount`, so this shouldn't happen, right?

To reproduce this, run the `run_beancount_import.py` script in a directory containing the following three files. (I am getting a lot of these in a real set of transactions but this example is made up.)

`in.ofx`:
```
<OFX>
 <BANKMSGSRSV1>
  <STMTTRNRS>
   <STMTRS>
    <CURDEF>CAD
    <BANKACCTFROM>
     <ACCTID>ACCT0
     <ACCTTYPE>CHECKING
    </BANKACCTFROM>
    <BANKTRANLIST>
     <STMTTRN>
      <TRNTYPE>CREDIT
      <DTPOSTED>20230101000000[-8:PST]
      <TRNAMT>20.00
      <FITID>0
      <NAME>Transaction 0
     </STMTTRN>
     <STMTTRN>
      <TRNTYPE>DEBIT
      <DTPOSTED>20230101000000[-8:PST]
      <TRNAMT>-10.00
      <FITID>1
      <NAME>Transaction 1
     </STMTTRN>
    </BANKTRANLIST>
   </STMTRS>
  </STMTTRNRS>
 </BANKMSGSRSV1>
</OFX>
```

`main.beancount`:
```
2021-06-23 open Assets:Chequing CAD
  ofx_org: ""
  ofx_broker_id: ""
  account_id: "ACCT0"
  ofx_account_type: "cash_only"
```

`run_beancount_import.py`:
```
import beancount_import.webserver
import os

journal_dir = os.path.dirname(__file__)

data_sources = [
    {
        "module": "beancount_import.source.ofx",
        "ofx_filenames": [f"{journal_dir}/in.ofx"]
    },
]

beancount_import.webserver.main(
    (),
    journal_input = os.path.join(journal_dir, "main.beancount"),
[...]

PR #206, opened 2023-09-03
fix: #202 - missing checks during matching for number

JonBoyleCoding:
8ccfee5 Changed the matching algorithm, and did not put in a check on the match_amount as had previously been done in _get_weight_matches.

This fix adds the check to ensure that the number is within the fuzzy range.

### DIFF
--- a/beancount_import/matching.py
+++ b/beancount_import/matching.py
@@ -456,6 +456,10 @@ class PostingDatabase(object):
                     if posting_date and posting_date != date:
                         continue
 
+                # Verify that number is in fuzzy range
+                if abs(sp.number - amount.number) > self.fuzzy_match_amount:
+                    continue
+
                 key = _entry_and_posting_ids_key(sp.entry, sp.mp)
                 matches[key] = (sp.entry, sp.mp)
         return matches


Zburatorul:
Good find. Could you add a unit test that would have caught the absence of this check?

ankurdave:
@JonBoyleCoding Thanks for the fix! Here's a unit test that previously failed and now passes:

```python
# In matching_test.py:

def test_nonmatch_fuzzy_amount_with_dates():
  # We don't match with a skew of 0.02, beyond our configured fuzzy_match_amount.
  assert_match(
      pending_candidate="""
      2023-01-01 * "Transaction 0"
        Assets:A 20.00 USD
          date: 2023-01-01
          cleared: TRUE
        Expenses:FIXME -20.00 USD
      """,
      pending="""
      2023-01-01 * "Transaction 1"
        Assets:B -19.98 USD
          date: 2023-01-01
          cleared: TRUE
        Expenses:FIXME 19.98 USD
      """)
```

ankurdave:
It seems like this is because of the indexing logic here: https://github.com/jbms/beancount-import/commit/8ccfee5b4ce662f8e070a90a9c0961769488391e#diff-1d4595bc7bf509394ca45262f4656c4b180c83a7be5e6cb653b7265d29743c97R445

We currently modify the lower bound of the loop to include an element less than the lower bound:
```python
for sp in cur_matches[lower_bound-1 if lower_bound > 0 else 0:upper_bound]:
```

Removing that fixes the problem:
```python
for sp in cur_matches[lower_bound:upper_bound]:
    posting = sp.mp.posting
    # Verify that number is in fuzzy range
    if abs(sp.number - amount.number) > self.fuzzy_match_amount:
        assert False  # this assertion is no longer hit
```

@xentac Do you remember why we modify `lower_bound` here?

ankurdave:
Opened https://github.com/jbms/beancount-import/pull/210 with the above test and fix.

JonBoyleCoding:
Sorry for not getting back with a test - I was partially unsure of how this should have been done and then work got in the way.

With the other above PRs, will this fix still be necessary and worth merging in? Or will they supercede the problem?

ankurdave:
@JonBoyleCoding Not at all! It did take me a little while to find the test, even with the repro in #202. It turned out that the `date` attribute was necessary to reproduce the bug.

#210 will supersede this PR.

JonBoyleCoding:
Closing as #210 is now merged and supercedes.

PR #220, opened 2023-12-31
reorder if statement to handle case where ticker is None

dkess:


jbms:
Duplicate of https://github.com/jbms/beancount-import/pull/244

PR #223, opened 2024-01-08
several minor OFX improvements

jktomer:
Bugs fixed:
* fix an assert failure when an ofx file included a zero-value transfer
* don't emit balance assertions for cash accounts (like we already don't for securities accounts) on the same day that there was a transaction

New features:
* support an additional after-tax account to be marked as related to a main account
* support listing multiple CUSIPs for the same security (for securities that don't have a real CUSIP and different banks assign different fake ones for their OFX feeds)
* support a configurable format string for producing the payee for a transaction
* support providing historical account IDs for an account


Zburatorul:
I noticed the ultipro PR also skips balance entries on same day as other transactions. Is there a way to handle this at a higher level for all sources simultaneously, with an optional flag to do this?

PR #234, opened 2024-04-21
Make generic importer more flexible

### LINKED ISSUE #204: GenericImporterSource - which data/field is used for the source_desc field?
I have written a GenericImporter for my bank statements and I am automatically setting the `payee` and `narration` based on my "raw" data / statements.

Sometimes I leave the `narration` empty, which seems to be a problem with beancount-importer. It seems that the narration field is used/copied to the source_desc field which is used for the recognition of already confirmed transactions.

As my bank statements have their "raw" data which I would like for the `source_desc` matching, I tried to add them straight as metadata, but they are getting overwritten by beancount-importer. 

How can i define what beancount-importer uses for the `source_desc` field?

### LINKED ISSUE #232: Does not properly clear transactions with multiple postings on the main account
With a `generic_importer_source`, if a transaction has multiple postings with the importer's main account (e.g. to split regular salary and bonus in a payslip), upon import only the first such posting will get the `source_desc` meta field to be cleared.
The logic to check for uncleared postings, however, then checks all such postings, and immediately marks it as uncleared, because some postings weren't cleared during import.

I then have to go in and manually copy over the meta field to the other posting(s).

dumbPy:
* support multiple postings from source account in generic importers
    closes #232
* allow importers to override source description
    closes #204

### DIFF
--- a/beancount_import/source/generic_importer_source.py
+++ b/beancount_import/source/generic_importer_source.py
@@ -93,8 +93,8 @@ class ImporterSource(DescriptionBasedSource):
         for i, posting in enumerate(postings):
             if posting.account != self.account: continue
             if isinstance(posting.meta, dict):
-                posting.meta["source_desc"] = entry.narration
-                posting.meta["date"] = entry.date
+                posting.meta.setdefault("source_desc", entry.narration)
+                posting.meta.setdefault("date", entry.date)
             else:
                 to_mutate.append(i)
         for i in to_mutate:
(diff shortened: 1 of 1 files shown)


Zburatorul:
I'd love to review this but not sure how to test. Could you add a sample configured source please, with some description of the parameters?

dumbPy:
@Zburatorul  Added example showing how to use the changes from this PR

Zburatorul:
Can you add an empty commit please? I can't figure out how to rerun the build workflow.

Issue #218, opened 2023-12-29
schwab_csv importer raises AssertionError because CSV has no title line


kerrickstaley:
## Summary
The CSV that I downloaded from Schwab at https://client.schwab.com/app/accounts/history doesn't have a `Transactions for account...` line, so I get an `AssertionError` when trying to import it.

## To reproduce
1. Download a CSV by going to https://client.schwab.com/app/accounts/history and clicking the download icon in the top right. The first line of the CSV should be:
    ```
    "Date","Action","Symbol","Description","Quantity","Price","Fees & Comm","Amount"
    ```
2. Put the CSV in your "transactions" directory that you pass to `transactions=glob(...)` in your config
3. Run beancount-import

## Expected behavior
No error

## Actual behavior
You get this exception:
```
Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/webserver.py", line 514, in _handle_reconciler_loaded
    loaded_reconciler = loaded_future.result()
                        ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/thread_helpers.py", line 13, in wrapper
    f.set_result(fn(*args, **kwargs))
                 ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/reconcile.py", line 380, in __init__
    self._load_sources()
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/reconcile.py", line 437, in _load_sources
    sources = self.sources = [
                             ^
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/reconcile.py", line 438, in <listcomp>
    load_source(spec, log_status=self.reconciler.log_status)
  File "/opt/homebrew/lib/python3.11/site-packages/beancount_import/source/__init__.py", line 334, in load_source
    return m.load(source_spec, log_status=log_status)  # type: ignore
[...]

Issue #171, opened 2022-08-03
Schwab brokerage transcations csv changes and unhandled BrokerageActions


pmeckoni:
I compared the schwab bokerage transactions csv downloaded on 2022-08-02 with those in folder [testdata/source/schwab_csv/test_basic/transactions/](https://github.com/jbms/beancount-import/tree/master/testdata/source/schwab_csv/test_basic/transactions/)

Below are the changes in csv file and the unhandled BrokerageActions that I encountered with version 1.3.5

### 1. Last empty column is dropped

The CSV transaction files from Schwab brokerage have dropped the last empty column. 

Old header

     "Date","Action","Symbol","Description","Quantity","Price","Fees & Comm","Amount",

New headers

     "Date","Action","Symbol","Description","Quantity","Price","Fees & Comm","Amount"

### 2. First line and Last line have changes

I am not sure if this is useful, but the first and last lines have changed

Old first line

     "Transactions  for account Brokerage XXXX-1234 as of 11/15/2020 18:00:21 ET"

New first line 

     "Transactions  for account Brokerage ...234 as of 08/03/2022 02:16:23 AM ET"

Old last line

      "Transactions Total,"","","","","","",-$9620.64

New last line

      "Transactions Total","","","","","","","-$9620.64",

### 3. Unhandled BrokerageAction

I experience the following unhandled BrokeageActions:
1. `Pr Yr Non Qual Div`
2. `Cash Merger`
3. `Cash Merger Adj`
4. `Pr Yr Lt C/G Rein`
5. `Pr Yr Long Term Cap Gain`
6. `Special Qual Div`
7. `Name Change`

naskooskov:
I've ran into the same issue with the csv format changing. Additional transaction types I've hit are:

Assigned
Wire Sent
Options Frwd Split

Happy to supply example csv contents if that would help.

PR #235, opened 2024-05-29
Update Schwab CSV to support latest format

1997cui:
1. #218 I had a similar issue, so instead of reading the account number from the first line, we read it from the file name, so **DO NOT change the default Download filename**
2. #171 This partially resolved this by changing the line format.
3. I added more transaction types to the best of my knowledge.

### DIFF
--- a/beancount_import/source/schwab_csv.py
+++ b/beancount_import/source/schwab_csv.py
@@ -171,6 +171,7 @@ class BrokerageAction(enum.Enum):
     MISC_CASH_ENTRY = "Misc Cash Entry"
     MONEYLINK_DEPOSIT = "MoneyLink Deposit"
     MONEYLINK_TRANSFER = "MoneyLink Transfer"
+    ATM_WITHDRAW = "ATM Withdrawal"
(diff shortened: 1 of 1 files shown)


carljm:
I no longer use the Schwab CSV importer (I found it too painful because Schwab changes their format too frequently), so I can't personally test this. But if you get it passing CI I'll be willing to merge it on the basis that it works for you.

sammyvancleve:
A little late to the party, but just tested this fix and it looks like Schwab has further changed their formatting. Specifically, formatting the date in the transactions header from MM/DD/YYYY to YYYY/MM/DD -- manually changing the header of the transactions CSV took no time and worked for me.

As for the positions CSV, this worked for me if I exported the CSV from 'Legacy View', but their new default CSV was incompatible. I might try to update this fix but considering how often they seem to change their formatting I don't know how useful it would be long-term.

carljm:
FWIW I think the Schwab csv importer should perhaps be abandoned/removed. I contributed it originally but I've found they change their CSV formats constantly and so I've stopped using it, maintaining it seems like more work than manually entering transactions. (Though I guess that depends how many transactions you have.)

PR #238, opened 2024-09-17
Upgrade Node 15.x to 18.x

Zburatorul:
Builds started failing for several reasons including Github action version deprecation, changes to mypy, and beancount 3 getting released.
These are the minimal changes I identified for the workflow to succeed.

Things that will need more work:
We need to decide if we want to take beancount 3 as a dependency or stick with <3. If we move to 3 we need to move from beancount.ingest to beangulp. An additional issue popped when I tried that which is that `journal_editor_test.py` was failing due to the `__automatic__` tag. Unclear what that has to do with beancount 3 but it is so.

There are some other mypy failures I had to suppress related to abstract classes. Threads online provide mixed advice and it seems the consensus is that it may be a mypy issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,6 @@
 name: Build
 
+
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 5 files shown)


jbms (approved):

jbms:
Thanks for these fixes.  It would be nice to support beancount v3 but that could be done separately. 

For `__automatic__` we will need to see what it means but probably it will just need to be excluded.

Issue #241, opened 2024-10-25
Pluggable/determinstic processing feature request|


DarrenRiedlinger:
Would you be open to a feature or PR that would provide hooks for user_defined deterministic processing of matched transactions, thereby skipping the UI review?  This would be similar to the need in #229 but extend it further.  

In my use case, I have tons of transactions that are easily/reliably handled by a deterministic approach, but which take significant time to review.  Many of these also end up as split transactions that aren't easily handled in the UI--but are easily handled in separate, deterministic code. However, they still benefit from using the source importers and existing/duplicate txn removal provided by this project.  I'm currently handling this by doing an initial pre-import step in my launch script. That pre-import script 1) directly instantiates the Reconciler object, 2) gets it's computed list of pending_entries, 3) does some custom matching/processing on those pending_entries, and 4) then writes just those transactions it matched/processed out to the ledger with appropriate postings and metadata.  I can then re-start everything using the normal web-ui script and any transactions I previously handled automatically get removed as existing transactions.  This works, but seems a bit inelegant.

That said, even if you're open to a PR but I'm the one writing it, I'm a bit lost how to best integrate that into the existing code, so would need some pointers unless someone else takes that up.  Ie. wrapping loaded_reconciler.pending_entries with a hook for some user-defined deterministic processing seems straightforward.  But I'm a bit lost on how the change staging is getting done within the journal_editor.py and how to cleanly handle directly writing the matched/custom processed transactions to the ledger without messing anything else up.  I'm currently just directly printing to the output file selected from reconciler.entry_file_selector using beancount.parser.printer.print_entry(entry), but not sure if that would play nice with the rest of the project code.  Thanks.

Zburatorul:
Hi @DarrenRiedlinger,
I would welcome such a feature at least because I too have many deterministic transactions. I am not familiar with the part of the logic you have questions on. I am happy to review PRs and otherwise test.

DarrenRiedlinger:
Thanks! I'll take a stab at it and submit a PR when I have a chance.

Zburatorul:
Can you share your branch in the meantime? I'd love to take a look.

DarrenRiedlinger:
[...]

PR #243, opened 2024-11-27
Added deterministic preprocessor functionality

DarrenRiedlinger:
Proposed solution for issue #241 allowing user-defined, deterministic preprocessing of imported sources.

Rather than having the pre-processing logic hard-coded within the reconcile.py, I settled on a minimal approach where reconcile.py is just very slightly modified with a hook to pass the LoadedReconciler instance to a user-defined preprocessor function.

That user-defined preprocessor function can be any user-supplied function.  It has full access to the LoadedReconciler state, so it can directly modify the list of pending_entries, and also has access to the JournalEditor class for writing entries to the appropriate ledger.  So it hopefully allows for many use-cases.

For simple use cases, I also created a separate BasePreprocessor class and a TransactionRuleMixin.  User-defined preprocessors that subclass from these can specify pre-processing rules in a dictionary-like fashion

E.g.

```
from beancount_import.preprocessor import BasePreprocessor, TransactionRuleMixin, PreprocessingRule

rules = [
    PreprocessingRule(
        source_name="venmo",
        transaction_narration="Tutoring",
        posting_meta={"venmo_description": "Tutoring"},
        target_account="Expenses:Tutoring",
        action="record",
    ),
]


class RuleBasedPreprocessor(TransactionRuleMixin, BasePreprocessor):
    def preprocess(self) -> None:
        self.apply_rules(rules)
```

And this can be passed to the webserver like so:
```
beancount_import.webserver.main(
    ...
    preprocessor=RuleBasedPreprocessor,
    )
```

Let me know your thoughts on this approach.
If I had React experience + time it would be nice to also allow creating PreprocessingRules based on the selected pending entry directly within the web-ui, but that's unfortunately beyond what i can handle at the moment.


Zburatorul:
Finally got to try this. Worked pretty well! The example you gave in the comment above was very helpful. It would have taken a lot longer to figure out how to set it up. 
Could you add an example folder in [here](https://github.com/jbms/beancount-import/tree/master/examples) or some text in that README?

I haven't reviewed the code itself yet. Tests are failing on `list[]` vs `typing.List[]`

DarrenRiedlinger:
Great. Yeah, traveling this weekend so might be a couple days till I get to
it, but happy to add a more prominent example.

Also, since I submitted, I noticed some issues with updating the
pending_data of the passed LoadedReconciler instance that has to do with
whether those attributes get passed by reference or value and then whether
the changes “stick” outside the Preprocessor. I think it doesn’t need a big
change to fix, but need to dig into what exactly is happening when I have a
moment.

Thanks

Zburatorul:
Did you end up sorting out the reference/copy issue?

PR #244, opened 2024-12-02
Fix ticker is None check

philipsd6:
Code would never get checked for None in the original order, because an exception would be raised when `startswith` was called on None. Swapping the order is the sensible solution.


jbms:
Thanks

PR #250, opened 2025-05-01
Some Ideas on how to fix gh build / mypy type errors

bkleinen:
I've been working with this module for about three years now (just with amazon source), and am currently facing new import errors with multiple shipment orders. In order to fix them, I would like to do some refactoring and need working tests, so I looked into the gh actions and tried to figure out why they started failing a while ago. The only explanation I found was that the mypy module was updated several times, and as there seems to be no version locking in tox the error could have been introduced by a mypy upgrade. I introduced version locking with pipenv, decided to go ahead with mypy and rather fix/ignore the errors, but still think version locking is immanent as beancount 3 was released this january.

This PR contains a lot of changes, it would be great to hear what you think about it, and if needed I can split it up and remove my personal taste stuff (like Makefile or testloop.sh).

- imho the current dependency management had no way to fix to specific versions
- replaced with pipenv
- added Makefile for local runs
- started to fix mypy errors in one file, ignore others
- add beautiful soup proxy to locate not Null checks for bs4 in one place: beancount_import/api_proxies/beautifulsoup.py
- add testloop.sh


bkleinen:
I've added a new gh workflow to test out the dependency management with pipenv: 
https://github.com/bkleinen/beancount-import/actions/workflows/test-pipenv.yml
- the other workflow(s) have not yet been fixed.

jbms:
Thanks for working on this 

What do you think about using uv instead?  I have switched a number of projects to use it and found it to be extremely good.

bkleinen:
> What do you think about using uv instead? I have switched a number of projects to use it and found it to be extremely good.

https://github.com/astral-sh/uv ? I had a brief look, seems pretty good, but also very new? 
My personal opinion is that I don't see why the 10-100% is needed (in a small project),
and I'm hesitant to switch to a new tool and would rather stick with the currently 
widely-used option. But then it doesn't really matter as long as the dependency versions can be locked.

bkleinen:
as this pr contains 3 different things (mypy fix, pipenv, and Makefile/testloop.sh) I've split it in three parts. 
1. mypy fix: new PR #251 
2. pipenv build: https://github.com/beancount-stuff/beancount-import/tree/build_with_pipenv (no pr as windows build is not working. )
3. my personal stuff: https://github.com/beancount-stuff/beancount-import/tree/my_tools

PR #256, opened 2025-09-13
Add support for beancount v3

### LINKED ISSUE #253: Beangulp compatible?
The docs state that [beancount_import.source.generic_importer_source](https://github.com/jbms/beancount-import/blob/master/beancount_import/source/generic_importer_source.py): imports from beancount.ingest.importer.ImporterProtocol subclass Importer, however, I've migrated my importers to use beangulp now, so now my importers are subclassing beangulp.Importer, so... is there any way to use these importers with beancount-import yet?

LalitMaganti:
Switch to using beangulp which supports both v2 and v3.

Fixes: https://github.com/jbms/beancount-import/issues/253

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ The currently supported set of data sources is:
 - [beancount_import.source.ultipro_google](beancount_import/source/ultipro_google.py):
   imports Google employee Ultipro payroll statements.
 - [beancount_import.source.generic_importer_source](beancount_import/source/generic_importer_source.py):
-  imports from `beancount.ingest.importer.ImporterProtocol` subclass Importers. See [beancount's documentation](https://beancount.github.io/docs/importing_external_data.html#the-importing-process) on how to write one and checkout the [examples](examples/) directory for a simple csv importer
+  imports from `beangulp.importer.ImporterProtocol` subclass Importers. See [beancount's documentation](https://beancount.github.io/docs/importing_external_data.html#the-importing-process) on how to write one and checkout the [examples](examples/) directory for a simple csv importer
 
 Refer to the individual data source documentation for details on configuration.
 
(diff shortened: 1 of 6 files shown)


Nulifier:
This doesn't appear to work for the `generic_importer_source`.

I get the following error:
```
TypeError: expected str, bytes or os.PathLike object, not _FileMemo
```

This is related to the switch from _FileMemo to straight filepath strings. I managed to get past that by checking if the filepath has a name property and using that. When I got past that, I ran into the following issue:
```
TypeError: Importer.extract() got an unexpected keyword argument 'existing_entries' which is related to the change in call signature from the Importer.
```

LalitMaganti:
Thanks for letting me know. https://github.com/jbms/beancount-import/pull/256/changes/0c37757e9cb8580e5b02244a2298867ebeb801ae should fix this.

slashformotion:
LGTM

Zburatorul:
@LalitMaganti thanks for kicking this off — I built on the same beangulp
  approach and opened #263 as a more complete alternative. It takes the same beangulp approach as this PR (the generic_importer_source change is equivalent), but also carries what's needed  for a complete v3 install:
  
  - adds `beangulp` and `beanquery` to `install_requires` (this PR switches the
    imports but doesn't declare the new deps);
  - migrates `delete_transactions`/`rename_account` off the removed
    `beancount.query` → `beanquery`;
  - handles v3's `__automatic__`/`__residual__` posting metadata in
    `journal_editor`; 
  - swaps `pkg_resources` → `importlib.resources`, fixes mypy under v3's
    `py.typed`, and updates CI to Python 3.9–3.12;
  - keeps v2.3+ compatibility (`beancount>=2.3.5`, since beangulp/beanquery
    both support 2.3+).
  
  CI is green across 3.9–3.12 on Linux/Windows/macOS and I validated it against a real ledger. Flagging here so we're not working the two in parallel — @jbms, happy to go with whichever you prefer.

